PR #160, opened 2026-04-06
fix: disable webkit2gtk DMABUF renderer on Linux to fix blank window

georgenijo:
## Summary

On Linux/Wayland with mesa or NVIDIA stacks (Fedora, Nobara, Ubuntu 23+), webkit2gtk's DMABUF renderer leaves Tauri windows rendered invisibly. The window chrome appears but the webview surface never paints — users see a blank window and conclude the app is broken.

Reproduced locally on Nobara (Fedora 43) / Wayland / NVIDIA: `./target/release/ui` rendered a blank dark window. Setting `WEBKIT_DISABLE_DMABUF_RENDERER=1` in the environment restored rendering immediately.

This PR bakes the workaround into the binary so Linux users get a working window out of the box, with an env-var opt-out for anyone whose stack renders correctly without it.

## Changes

- `app/src-tauri/src/lib.rs` — Sets `WEBKIT_DISABLE_DMABUF_RENDERER=1` and `WEBKIT_DISABLE_COMPOSITING_MODE=1` at the top of `run()`, cfg-gated to Linux. Only applied if the variable isn't already set, so users can opt out or customize. The logic lives in `apply_linux_webkit_env_defaults()` with injected get/set closures, letting the precedence rules be unit tested against a fake env map.
- 4 new unit tests covering: empty env, user overrides preserved, partial defaults, empty-string treated as set.
- `.github/workflows/release.yml` — New smoke test step in `release-linux` that launches the built binary under `xvfb-run` + `dbus-run-session` and waits up to 20s for `[main] App mounted` in the telemetry log. Fails the release if the frontend never mounts — catches rendering regressions before they ship.

## Test plan

- [x] `cargo test --lib` — 70 passed (4 new)
- [x] Locally: `./target/release/ui` renders window + mounts React on Nobara Wayland without any env vars
- [ ] CI: `rust-linux` job runs the new unit tests on ubuntu
- [ ] CI: `release-linux` smoke test runs on next tag push

## Notes

macOS behavior is untouched — the entire apply function and tests are `#[cfg(target_os = "linux")]`.

The smoke test is the first automated launch verification for Linux. It mirrors the existing macOS smoke test (release.yml:87) but adds a stronger assertion (frontend mount vs just process-alive).

## Summary by CodeRabbit

* **Tests**
  * Added automated Linux smoke tests for release builds that validate app startup, initialization, and mount signal detection, with detailed logging on failure.

* **Chores**
  * Applied Linux-specific defaults to improve web-rendering and startup stability on Linux builds.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -208,6 +208,56 @@ jobs:
         with:
           projectPath: app
 
+      - name: Install smoke test dependencies
+        run: sudo apt-get install -y xvfb dbus-x11 at-spi2-core
+
+      - name: Smoke test — verify binary launches and frontend mounts
+        run: |
+          BINARY="app/src-tauri/target/release/ui"
+          LOG="$HOME/.local/share/local-dictation/logs/app.log"
+
+          echo "=== Verifying binary exists ==="
+          ls -la "$BINARY"
+
+          echo "=== Clearing any prior log ==="
+          rm -f "$LOG"
+
+          echo "=== Launching under xvfb + dbus ==="
+          xvfb-run -a --server-args="-screen 0 1280x800x24" \
+            dbus-run-session -- "$BINARY" > /tmp/murmur-smoke.log 2>&1 &
+          PID=$!
+          echo "Started with PID $PID"
+
+          echo "=== Waiting up to 20s for [main] App mounted ==="
+          MOUNTED=0
+          for i in $(seq 1 20); do
+            sleep 1
+            if ! kill -0 "$PID" 2>/dev/null; then
+              echo "FAIL: Process died after ${i}s"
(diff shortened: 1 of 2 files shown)


georgenijo:
Addressed CodeRabbit review feedback (1 actionable comment).

**Fixed:**
- `app/src-tauri/src/lib.rs` `treats_empty_string_as_set` test — replaced the loose `assert!(!writes.contains(...))` check with an exact `assert_eq!` against the expected writes vector (`["WEBKIT_DISABLE_COMPOSITING_MODE"]`). The previous assertion would have passed even if `set` was never called; the new one verifies the compositing-mode default is still written when `WEBKIT_DISABLE_DMABUF_RENDERER` is present as an empty string. All 70 unit tests pass.

No other actionable items were raised in the review.

PR #162, opened 2026-04-20
feat: Add diagnostic logging to keyboard module for silent rejection and tap health

### LINKED ISSUE #152: Add diagnostic logging to keyboard module for silent rejection and tap health
## Context

Investigation of #137 across three independent analyses of production logs (115k lines, 25 days) found **zero confirmed missed hotkey events** — every keyboard emission was received and processed by the frontend. But the current logging only captures *successful* detections, making it impossible to diagnose the two most likely root causes:

1. **rdev LAST_FLAGS desync** — modifier press/release is synthesized by diffing cached flags against incoming `FlagsChanged` events. If macOS suppresses a transition, the next tap arrives as the wrong edge or `None` and is silently dropped before Murmur's state machine sees it.
2. **Failed double-tap timing** — the 200ms hold limit and 400ms gap window silently reject taps with no log entry. Users perceive this as "hotkey not working."

### 1. Raw rdev callback logging (trace level)
Log every event the rdev callback receives: event type, key, current detector state, and the inferred press/release result for `FlagsChanged` events. This is the only way to distinguish "rdev didn't fire" from "rdev fired but was rejected."

### 2. State machine rejection reasons
When the double-tap or hold-down detector resets without firing, log the reason:
- `held_too_long` — first press exceeded 200ms
- `second_tap_expired` — gap exceeded 400ms
- `combo_cancelled` — non-modifier key pressed during sequence
- `single_short_tap_noop` — Both-mode short tap, intentionally no-op
- `processing_skipped` — IS_PROCESSING was true, event dropped

### 3. Tap silence heartbeat
Extend the existing heartbeat thread: track "last raw rdev callback seen at" and log a warning if no events arrive for an extended period (e.g., 5 minutes) while the listener is active. This would surface CGEventTap timeout or Secure Input suppression.

### 4. set_processing caller logging
Log every `set_processing(true)` and `set_processing(false)` call with the caller context, so we can detect stuck-true scenarios.

## Files
- `app/src-tauri/src/keyboard.rs` — all changes here

## Prerequisite for
- rdev LAST_FLAGS resilience work
- UX feedback for failed taps

georgenijo:
Closes #152

## Pipeline Run
- Plan: plan-v2.md
- Review rounds: 0
- Test rounds: 0
- Status: All tests passing

Logs: `pipeline-logs/murmur-app/issue-152/`

## Summary by CodeRabbit

* **New Features**
  * Keyboard listener now captures detailed diagnostic information on state transitions and event handling for improved troubleshooting
  * Logging verbosity can be customized via the RUST_LOG environment variable while maintaining sensible defaults
  * Added heartbeat monitoring to detect listener inactivity

* **Tests**
  * Added validation for keyboard listener diagnostic capture functionality

### DIFF
--- a/app/src-tauri/src/keyboard.rs
+++ b/app/src-tauri/src/keyboard.rs
@@ -16,7 +16,7 @@ use rdev::{listen, Event, EventType, Key};
 #[cfg(target_os = "macos")]
 use rdev::set_is_main_thread;
 use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
-use std::sync::Mutex;
+use std::sync::{Mutex, OnceLock};
 use std::time::Instant;
 use tauri::Emitter;
 
@@ -31,6 +31,10 @@ const COOLDOWN_MS: u128 = 50;
 
 /// Cooldown after hold-down stop to prevent accidental re-trigger
 const HOLD_DOWN_COOLDOWN_MS: u128 = 50;
+
+/// If no raw rdev callback has fired for this long while the listener is
+/// active, emit a warn-level log. Tune down temporarily when verifying.
+const SILENCE_WARN_MS: u64 = 5 * 60 * 1000;
 
 #[derive(Debug, Clone, Copy, PartialEq)]
 enum DetectorState {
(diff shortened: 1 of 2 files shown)


georgenijo:
Issue #152 was completed via #174 (merged May 27). This PR was never merged and is stale. Closing.

PR #165, opened 2026-04-20
fix: rdev CGEventTap timeout recovery and LAST_FLAGS resilience

### LINKED ISSUE #153: rdev CGEventTap timeout recovery and LAST_FLAGS resilience
## Context

Three independent investigations of #137 converged on two rdev-level failure modes that would cause hotkeys to silently stop working:

### CGEventTap timeout
macOS silently disables a CGEventTap if its callback blocks for >1 second (`kCGEventTapDisabledByTimeout`). The pinned rdev build has **no handler** for this event type — the tap stays "alive" (no error logged) but stops receiving events entirely. This is standard practice to handle for any CGEventTap consumer.

### LAST_FLAGS desync
rdev synthesizes modifier KeyPress/KeyRelease by diffing a cached `LAST_FLAGS` value against incoming `FlagsChanged` events (`common.rs:74`). Murmur only allows modifier hotkeys (`keyboard.rs:319`), making it uniquely vulnerable. If macOS suppresses one `FlagsChanged` transition (sleep/wake, Secure Input toggle, Mission Control, full-screen space switch), the cached flags go stale and subsequent modifier events arrive as the wrong edge or `None` — silently dropped.

### CGEventTap timeout
Either:
- Fork/patch rdev to detect `kCGEventTapDisabledByTimeout` and call `CGEventTapEnable(tap, true)` to re-enable
- Add a periodic health check in Murmur's heartbeat thread that verifies the tap is still enabled and re-enables if needed

### LAST_FLAGS desync
Either:
- Patch rdev's `FlagsChanged` handler to be more robust to missed transitions (e.g., use absolute flag state instead of diffing against cached)
- Add a periodic flags resync in Murmur (e.g., on `NSWorkspaceDidWakeNotification`, or on a timer)
- As a workaround, reset the keyboard listener on wake events

## Depends on
- #152 (diagnostic logging) to confirm these are the actual root causes before investing in fixes

## Related
- #137

georgenijo:
Closes #153

## Pipeline Run
- Plan approved through full review loop
- Builder committed implementation (commit 3e62891)
- Tester identified 4 issues before watchdog timeout (see logs)
- **Known issues requiring manual fix:**
  - Cargo.lock not updated (BLOCKING)
  - rdev uses local path instead of git+rev (BLOCKING)
  - WAKE_OBSERVER raw-pointer Send/Sync issue (HIGH)
  - Missing restart_listener_full_sequence test (MEDIUM)

Logs: `pipeline-logs/murmur-app/issue-153/`

## Summary by CodeRabbit

* **New Features**
  * Added manual keyboard listener reset functionality.

* **Bug Fixes**
  * Improved keyboard listener reliability and recovery after system sleep (macOS).
  * Enhanced detection and handling of keyboard listener interruptions.

### DIFF
--- a/app/src-tauri/Cargo.toml
+++ b/app/src-tauri/Cargo.toml
@@ -28,7 +28,13 @@ arboard = "3"
 hound = "3.5"
 dirs = "5"
 cpal = "0.15"
-rdev = { git = "https://github.com/Narsil/rdev", branch = "main" }
+# Forked for issue #153 — see app/src-tauri/src/keyboard.rs header comment.
+# Patches: (1) re-enable CGEventTap on kCGEventTapDisabledByTimeout/ByUserInput
+# and expose tap/runloop/stopped handles (macos_reenable_tap, macos_is_tap_enabled,
+# macos_listener_stopped, stop); (2) absolute-flag-state modifier resolution with
+# equal-flags resync heuristic to survive missed FlagsChanged transitions.
+# Apple docs: CGEventTapEnable, CGEventTapIsEnabled, CFRunLoopStop.
+rdev = { path = "../../../../rdev" }
 memory-stats = "1"
 reqwest = { version = "0.12", default-features = false, features = ["stream", "rustls-tls"] }
 futures-util = { version = "0.3", default-features = false, features = ["alloc"] }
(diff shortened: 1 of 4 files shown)


georgenijo:
Superseded by #194's rdev-fork approach (which also supersedes #153). Closing this older app-side attempt; continue hotkey reliability work under #194.

PR #166, opened 2026-04-20
feat: surface amber overlay pulse when double-tap timing is rejected (#154)

### LINKED ISSUE #154: Surface feedback when hotkey tap is rejected (timing miss)
## Problem

The double-tap detector has a 200ms hold limit and 400ms gap window. Both-mode intentionally emits nothing for a short single tap. When a user's timing is slightly off, absolutely nothing happens — no visual, no sound, no log. This makes timing misses indistinguishable from "the app is broken."

Three independent investigations of #137 identified this as one of the two most likely explanations for the reported bug.

## Proposal

When the keyboard state machine rejects a tap attempt, emit a lightweight signal the user can perceive:
- Brief overlay flash or color change (e.g., overlay border flickers red/amber for 200ms)
- Optional subtle audio cue
- The feedback should be distinct from the "recording started" signal so it reads as "I saw your tap but it didn't qualify"

This should be gated behind a setting (off by default) since experienced users who understand the timing won't want the noise.

## Depends on
- #152 (diagnostic logging adds the rejection-reason tracking this feature would surface)

## Related
- #137

georgenijo:
Closes #154

## Pipeline Run
- Plan approved through full review loop
- Builder committed implementation (commit 163f929)
- Builder was killed by watchdog after completion (known cmux-hook hang)
- Tester/CI not run by pipeline — needs manual validation

Logs: `pipeline-logs/murmur-app/issue-154/`

### DIFF
--- a/app/src-tauri/src/keyboard/detectors.rs
+++ b/app/src-tauri/src/keyboard/detectors.rs
@@ -1,0 +1,1625 @@
+//! Keyboard detector types and logic (platform-neutral).
+//!
+//! Contains the local `Key` and `EventType` enums, both detector structs
+//! (`DoubleTapDetector`, `HoldDownDetector`), all shared global statics, and
+//! the `handle_event` dispatch function used by the listener thread.
+
+use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
+use std::sync::Mutex;
+use std::time::Instant;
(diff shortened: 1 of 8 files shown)


georgenijo:
Closing as superseded by merged PR #226, which implemented and fully verified the same #154 behavior on current main.

PR #182, opened 2026-05-28
fix: check mic permission natively instead of getUserMedia (#177)

### LINKED ISSUE #177: Bug: app audio quiets or ducks when Murmur focus changes
## Problem

Murmur appears to trigger unwanted audio quietening/ducking behavior related to app focus. When Murmur is out of focus, or when focus changes around recording/keyboard-triggered dictation, other app/system audio may become noticeably quieter.

This is distinct from hotkeys not firing when the app is unfocused (#137). Here the concern is that Murmur may be changing or causing macOS audio behavior in a way that affects other audio playback.

## Expected behavior

Murmur should not reduce, duck, or otherwise alter other apps' audio volume just because:

- Murmur is focused or unfocused
- the user starts/stops recording
- the user triggers recording via hold-down/double-tap
- the main window is hidden while Murmur continues running

## Actual behavior

Other audio can become quieter/ducked when Murmur focus changes or when Murmur is used from the background.

## Areas to investigate

- macOS input/audio session behavior during `cpal` recording
- Whether opening/closing the input stream changes system audio routing or activates voice-processing style behavior
- Whether the selected microphone device, Bluetooth/AirPods, or system default input/output pairing affects the ducking
- Whether the behavior happens only while recording or persists after stop
- Whether the main Tauri window focus state matters, or if it is purely tied to native audio capture

## Suggested diagnostics

- Reproduce while playing audio from another app, then start/stop Murmur recording while Murmur is focused and unfocused.
- Compare built-in mic vs external USB mic vs AirPods/Bluetooth mic.
- Check Murmur structured logs around recording start/stop, selected input device, and audio stream lifecycle.
- Verify whether `cpal` exposes any macOS-specific stream configuration that could avoid ducking.

## Acceptance criteria

- Root cause is identified and documented.
- Murmur no longer causes other app/system audio to quiet unexpectedly during focus changes or recording lifecycle, if avoidable.
- If this is a macOS/Bluetooth limitation that cannot be fully prevented, the limitation is documented and the app gives users a clear workaround.
- Native macOS smoke test verifies audio playback volume remains stable while Murmur records from the background and foreground.

georgenijo:
Fixes #177.

## Root cause
`PermissionsBanner` checked microphone permission with `navigator.mediaDevices.getUserMedia({ audio: true })` on **every window focus**. In WKWebView that opens the mic through macOS voice-processing I/O (VPIO), which **ducks all other system audio** while active. So bringing Murmur to the foreground quieted/stuttered audio playing in other apps — on every focus, even after permission was granted (the component stays mounted and keeps a `focus` listener that re-probes).

Not Bluetooth/HFP, not the overlay, not recording — confirmed by objectively measuring system output level (RMS) across surface/leave cycles. Stubbing out the `getUserMedia` call eliminated the duck; the native fix reproduces that.

## Fix
- `check_microphone_permission` Tauri command reads `AVCaptureDevice.authorizationStatus(for: .audio)` via objc2 — queries TCC state, never opens the device, so no VPIO and no ducking.
- `PermissionsBanner` uses the command instead of `getUserMedia`.
- `build.rs` links AVFoundation for the `AVCaptureDevice` lookup.

## Tests
- **CI guard** `app/src/lib/no-mic-probe.test.ts` — fails the build if `getUserMedia(` is reintroduced anywhere in the frontend (the WebView must never open the mic; recording is native via cpal).
- **Manual QA harness** `scripts/qa/` — `rms-meter.swift` taps system output level; `audio-duck-check.sh` scripts a surface/leave cycle and reports PASS/FAIL on whether foregrounding ducks other audio. Verified PASS on this fix, FAIL on the prior behaviour.

## Verification
- meter/harness: PASS (no duck) on fixed build
- `npm test` 30 pass · `cargo test` 84 pass · `tsc` clean

> Stacked on #181 (vitest localStorage env fix) so CI is green; retarget to `main` after #181 merges.


georgenijo:
@coderabbitai review

PR #184, opened 2026-05-29
feat: add macOS Accessibility permission reset button

### LINKED ISSUE #176: Add a macOS Accessibility permission reset button
## Problem

Murmur can get into a confusing macOS Accessibility permission state, especially for rebuilt/dev Tauri app bundles: System Settings may show the app in Privacy & Security > Accessibility, but the running app still reports that Accessibility access is missing.

We hit this while testing `Local Dictation Dev`: granting the visible Accessibility entry was not enough until the stale TCC entry was reset and the app was re-enabled.

## Proposal

Add a secondary troubleshooting action in the app for Accessibility permission failures.

Suggested UX:

- Keep the primary `Grant` / `Open Settings` flow for the normal case.
- Add a secondary `Reset Permission` or `Reset & Open Settings` button when Accessibility permission is still missing.
- The button should reset Murmur's Accessibility TCC entry for the current bundle identifier, then open the macOS Accessibility settings pane.
- Copy should make clear that macOS still requires the user to turn the app back on manually.

## Implementation notes

- Do not reset all Accessibility permissions globally.
- Reset only the current bundle identifier, e.g. via `tccutil reset Accessibility <bundle-id>`.
- The app cannot grant itself Accessibility permission; this button can only clean up a stale entry and send the user to the right place.
- Useful locations:
  - `app/src-tauri/src/commands/permissions.rs`
  - `app/src/components/PermissionsBanner.tsx`
  - `app/src/components/settings/SettingsPanel.tsx`

## Acceptance criteria

- When Accessibility is missing, the user has a visible troubleshooting action to reset this app's stale Accessibility entry.
- The action opens the correct macOS System Settings pane afterward.
- The reset targets only Murmur's current bundle identifier.
- UI does not imply one-click permission granting.
- Existing checks pass:
  - `cd app/src-tauri && cargo check`
  - `cd app/src-tauri && cargo test -- --test-threads=1`
- Native smoke test verifies the button appears and the normal Accessibility flow remains understandable.

georgenijo:
Closes #176

## What

Adds a troubleshooting action for the confusing macOS Accessibility state where System Settings lists the app under Privacy & Security > Accessibility, but the running build still reports access missing (common after rebuilding a dev Tauri bundle, e.g. `Local Dictation Dev`).

## Changes

- **`commands/permissions.rs`**: new `reset_accessibility_permission` command. Reads the *runtime* bundle id via `NSBundle.mainBundle.bundleIdentifier` (not the static `tauri.conf.json` value, which is stale for dev bundles), runs `tccutil reset Accessibility <bundle-id>`, then reopens the Accessibility pane via the existing `open_system_preference_pane` helper. macOS-only; errors surfaced to the UI.
- **`lib.rs`**: registers the new command.
- **`PermissionsBanner.tsx`**: secondary "Still not working? Reset & Open Settings" button shown only when Accessibility is missing, plus helper copy clarifying the entry is cleared and the user must re-enable Murmur manually. Refreshes permission status after invoking.
- **`CHANGELOG.md`**: entry under Unreleased > Added.

## Acceptance criteria

- Visible troubleshooting action when Accessibility is missing
- Opens the correct System Settings pane afterward
- Resets only Murmur's current bundle id (never all apps)
- UI copy does not imply one-click granting

## Verification

- `cd app/src-tauri && cargo check` — clean
- `cd app/src-tauri && cargo test -- --test-threads=1` — 64 passed
- `cd app && npx tsc --noEmit` — clean
- Native smoke test left to manual QA (running it live would `tccutil reset` the real machine's permission)

## Summary by CodeRabbit

* **New Features**
  * Added macOS Accessibility troubleshooting: A new "Reset & Open Settings" button appears in the permissions banner when Accessibility is not granted, allowing users to reset stale permission entries and reopen System Settings to re-enable access.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Added
+- **Accessibility permission reset** troubleshooting button in the permissions banner — resets the app's stale TCC entry for the current bundle id (`tccutil reset Accessibility`) and reopens System Settings
+
 ### Fixed
 - Strip recording-status-changed emissions from `ensure_vad_model` to reduce event noise
 
--- a/app/src-tauri/Cargo.lock
+++ b/app/src-tauri/Cargo.lock
@@ -5746,7 +5746,7 @@ dependencies = [
 
 [[package]]
 name = "ui"
-version = "0.9.0"
+version = "0.9.1"
 dependencies = [
  "arboard",
  "base64 0.22.1",
--- a/app/src-tauri/src/commands/permissions.rs
+++ b/app/src-tauri/src/commands/permissions.rs
@@ -41,6 +41,80 @@ pub fn request_accessibility_permission() -> Result<(), String> {
     { Ok(()) }
 }
 
+/// Read the running process's bundle identifier (macOS).
+///
+/// Returns the *runtime* bundle id (e.g. the dev bundle `Local Dictation Dev`),
+/// which is what TCC actually keys Accessibility entries on — not the static
+/// identifier from `tauri.conf.json`, which can be stale for rebuilt/dev bundles.
+#[cfg(target_os = "macos")]
+fn current_bundle_identifier() -> Option<String> {
+    use objc2::msg_send;
+    use objc2::runtime::{AnyClass, AnyObject};
+    use objc2_foundation::NSString;
+
+    unsafe {
+        let cls = AnyClass::get(c"NSBundle")?;
+        let bundle: *mut AnyObject = msg_send![cls, mainBundle];
+        if bundle.is_null() {
+            return None;
+        }
+        let ident: *const NSString = msg_send![bundle, bundleIdentifier];
+        if ident.is_null() {
+            return None;
+        }
+        let s = (*ident).to_string();
+        if s.is_empty() {
+            None
+        } else {
+            Some(s)
+        }
+    }
+}
+
+/// Reset this app's stale macOS Accessibility TCC entry, then reopen the pane.
+///
+/// Troubleshooting action for the case where System Settings lists the app under
+/// Accessibility but the running build still reports access missing (common after
+/// rebuilding a dev bundle). Resets ONLY the current bundle identifier via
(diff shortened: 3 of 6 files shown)


georgenijo:
Thanks for the review. Addressed in the latest commit:

**1. Surface reset failures in the banner (Major) — fixed.**
`handleResetAccessibility` now sets a local `resetError` state on failure and renders an inline red message in the banner; it's cleared at the start of each attempt and `checkPermissions()` still runs in `finally`.

**2. Route the command through a `lib/dictation.ts` wrapper (Refactor) — declining.**
`lib/dictation.ts` is scoped to the dictation pipeline (init/start/stop/configure/transcribe). Every sibling permission command invokes Tauri directly — `check_accessibility_permission`, `check_microphone_permission`, `request_accessibility_permission`, `request_microphone_permission` in `PermissionsBanner.tsx`, and `request_accessibility_permission` in `SettingsPanel.tsx`. There is no permissions wrapper module. Adding a single wrapper for the reset command in a dictation-pipeline file, while its four siblings stay direct, would make the codebase less consistent, not more. Keeping the direct `invoke` matches the established convention for permission commands. Happy to extract a dedicated `lib/permissions.ts` for all of them as a separate follow-up if preferred.

georgenijo:
Reversed my earlier call on finding #2 — applied it. Added `resetAccessibilityPermission()` wrapper in `lib/dictation.ts` and `PermissionsBanner.tsx` now calls the helper instead of invoking the Tauri command directly. tsc clean. Commit b8f2b27.

georgenijo:
Correction to my previous comment: the wrapper export did not land in that commit (5b67932), which broke the build (tsc TS2305). Fixed in 4d9a8c9 — `resetAccessibilityPermission()` is now exported from `lib/dictation.ts` and tsc passes clean. Apologies for the premature 'tsc clean' note.

georgenijo:
Hash correction: the build fix is commit `87324b1` (not `4d9a8c9` as I mistyped above). Verified: tsc passes, `resetAccessibilityPermission` exported from `lib/dictation.ts` and consumed in `PermissionsBanner.tsx`.

PR #138, opened 2026-03-16
feat: overlay hover-expand with quick settings panel

georgenijo:
Closes #135

## Summary
- Hovering the Dynamic Island overlay pill expands it downward to reveal a compact row of quick-access controls: global disable toggle, auto-paste toggle, and gear icon to open settings
- Pre-allocates a taller overlay window with transparent overflow for flicker-free CSS height transitions
- Adds `disabled` field to Settings schema and `show_main_window` Rust command
- Cross-window settings sync via `settings-changed` Tauri event so overlay and main window stay in sync
- Inline recording timer (M:SS) appears in the top bar during recording + hover
- 300ms hover-intent delay on mouse-leave prevents flicker; controls fade in 100ms after expansion starts

## Test plan
- [ ] Hover over the overlay pill — verify it expands downward with a smooth animation
- [ ] Mouse-leave — verify it collapses after ~300ms delay
- [ ] Click speaker-slash icon — verify it toggles red disabled state; mic icon dims to 15% opacity
- [ ] Click auto-paste toggle — verify it toggles; verify main window settings panel reflects the change
- [ ] Click gear icon — verify main settings window opens and receives focus
- [ ] While disabled, verify auto-paste button is dimmed at 35% opacity and non-interactive
- [ ] Start recording via hotkey, then hover — verify red dot + inline timer + waveform in top bar
- [ ] Verify click/double-click on the pill (outside dropdown) still starts/stops recording
- [ ] Verify dropdown buttons don't trigger recording start/stop (stopPropagation)
- [ ] Test on external monitor (no notch) — verify fallback dimensions work

## Release Notes

* **New Features**
  * Overlay controls now appear on hover, enabling quick access to disable, auto-paste toggle, and settings.
  * Recording duration displays while recording and hovering over the overlay.
  * Settings are automatically saved and persist between sessions.
  * Improved overlay window sizing for better visibility.

PR #189, opened 2026-05-29
feat: overlay hover-expand with quick settings panel (dynamic resize)

### LINKED ISSUE #135: Overlay hover-expand with quick settings panel
## Summary

Add a hover-to-expand interaction to the Dynamic Island overlay. When the user hovers over the notch pill, it expands downward to reveal a compact row of quick-access controls. Mouse-leave collapses it back.

## Design

Figma file with all states: https://www.figma.com/design/ORegdcWBNhxjbwis1nwjuB?node-id=7-2

HTML mockup at `app/overlay-mockup.html` (temporary, can be deleted after implementation).

### States

**Idle (no hover):** Current behavior — compact pill with mic icon, width = `notchWidth + 28`.

**Idle (hover):** Pill expands downward. Top bar stays the same (mic icon). Below it, a single row appears with three controls:
- **Speaker-slash icon** — global disable toggle. Taps to disable Murmur (hotkeys stop working, recording can't start). See linked issue for backend wiring.
- **Auto-paste toggle** — toggle switch, reads/writes the `autoPaste` setting.
- **Gear icon** — opens the main settings window via `WebviewWindow.getByLabel('main')`.

**Recording (no hover):** Current behavior — expanded pill with red dot + waveform.

**Recording (hover):** Same dropdown row as idle hover. Top bar shows red dot + inline timer ("0:03") next to the waveform. No separate recording info pane — the red dot already communicates recording state.

**Disabled state:** When global disable is active, speaker-slash icon turns red (`rgba(239,68,68,0.12)` background, `#ef4444` stroke). Auto-paste toggle dims to 35% opacity. Top-bar mic icon fades to 15% opacity. Gear icon remains interactive.

### Key design decisions
- **Dropdown is identical regardless of state** — idle or recording, same controls. Only the top bar changes.
- **No model selector** — can't change model without a picker UI, not worth the space.
- **No recording mode selector** — same reasoning, needs a multi-option picker.
- **Timer is inline** — during recording+hover, "0:03" appears as small text next to the red dot in the top bar. No dedicated pane.
- **Speaker-slash icon** (not mic-slash, not power) — chosen because the top bar already has a mic icon; doubling up on mic shapes was confusing. Speaker-slash is universally "mute/disable" and visually distinct.

### Window sizing
The overlay window in `tauri.conf.json` is currently 260x100. The expanded state needs more height. Two approaches:
1. **Pre-allocate a taller window** (e.g., 260x160) and keep the extra space transparent/hidden until hover. Simpler, no resize flicker.
2. **Dynamically resize** on hover via `window.setSize()`. More precise but may cause visual jank.

Recommend option 1 — the transparent overflow is invisible on macOS.

### Hover detection
- `onMouseEnter` on the overlay surface triggers expansion.
- `onMouseLeave` collapses with a **300ms delay** (standard hover-intent pattern to prevent flicker during cursor movement).
- The overlay is `focusable: false` and uses `_setPreventsActivation` — mouse-enter/leave events don't require focus, so this should work. Verify in testing.

### Settings sync (the tricky part)
The overlay runs in a separate Tauri window with **no shared React context**. Today it already reads `localStorage` directly for the microphone setting (see `OverlayWidget.tsx` lines 188-197). The quick settings need to both read AND write:

- **Auto-paste toggle:** Read `autoPaste` from localStorage on hover-expand. Write back on toggle. Emit a custom event (`settings-changed`) so the main window picks up the change without a restart.
- **Global disable:** Read/write a new `disabled` boolean. The main window and keyboard listener need to respect this. See linked backend issue.
- **Settings window:** `invoke` or use `WebviewWindow.getByLabel('main')` + `show()` + `setFocus()`.

### Animation
- Expand/collapse: CSS transition on height, same spring curve as the existing width transition: `cubic-bezier(0.34, 1.56, 0.64, 1)` over 300-400ms.
- Controls fade in with a slight delay (100ms after height starts expanding) to avoid content popping.

### Files to modify
- `app/src/components/OverlayWidget.tsx` — hover state, expanded content, settings read/write
- `app/src-tauri/tauri.conf.json` — increase overlay window height if using pre-allocated approach
- `app/src/overlay.tsx` — may need to pass down expanded state
- `app/src/lib/settings.ts` — add `disabled` field to settings type and defaults

### What NOT to implement in this ticket
- The actual global disable backend wiring (gating hotkeys, blocking recording) — that's a separate issue.
- Haptic feedback on hover — deferred, requires private macOS API (`MultitouchSupport.framework`).

## Depends on
- Global disable backend wiring (linked issue) for the speaker-slash toggle to actually do something. The UI toggle can land first as a no-op, then get wired up.

georgenijo:
Closes #135

Alternative implementation to #138, rebuilt on current `main` (0.9.1). #138 is left untouched for comparison.

## Key differences vs #138
- **Dynamic window resize on hover** (`set_overlay_expanded`) instead of a pre-allocated tall window. A transparent Tauri window with cursor events enabled captures clicks across its whole frame, so a permanently-tall window would create a click dead-zone below the notch when idle. Grow window → animate open; animate closed → shrink window so the dropdown is never clipped.
- **Reuses the already-merged global-disable backend** (`set_app_disabled` / `APP_DISABLED`) — speaker-slash gates hotkeys + recording end-to-end (no settings.ts change needed; `disabled` already exists).
- **Gear opens the actual Settings panel** via an `open-settings` event, not just showing the window (panel open state is local React state in App.tsx).
- **Two-way `settings-changed` sync** with a diff-guard so a window applying its own change is a no-op (no echo loop); an expanded overlay reflects main-window changes live.
- **Drag region scoped to the top bar** so the dropdown buttons stay clickable.
- Inline `m:ss` recording timer in the top bar during recording + hover.

## Files
- `commands/overlay.rs` + `lib.rs` — `set_overlay_expanded` command.
- `OverlayWidget.tsx` — hover-intent expand/collapse, dropdown controls, timer, cross-window sync.
- `useSettings.ts` (+`applyExternalSettings`), new `useOverlaySettingsSync` / `useOpenSettingsListener` hooks, `dictation.ts` `buildConfigureOptions` helper.
- `docs/features/overlay.md`, `CHANGELOG.md`.

## Verification
- `cargo check` clean (no warnings); `cargo test -- --test-threads=1` → 84 passed.
- `tsc --noEmit` clean.
- Rendered the overlay dev entry and confirmed the dropdown layout + auto-paste toggle on/off.

Native smoke on a notched Mac still recommended: hover expand (no flicker / no click dead-zone when collapsed), gear opens Settings, speaker-slash gates hotkeys, auto-paste syncs both ways, inline timer while recording.

## Release Notes

* **New Features**
  * Hover-expand overlay quick-settings dropdown with global disable toggle, auto-paste control, and direct settings access.
  * Inline recording timer displayed in overlay during active recording sessions.
  * Warning banner alerting users when auto-paste is paused due to active file saving.

* **Documentation**
  * Updated overlay feature documentation with new interaction behavior and settings controls.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Added
+- **Overlay hover-expand quick settings** — hovering the Dynamic Island reveals a quick-settings dropdown with global-disable, auto-paste, and settings-window controls; inline recording timer while hovering (#135)
+
 ### Fixed
 - Strip recording-status-changed emissions from `ensure_vad_model` to reduce event noise
 
--- a/app/src-tauri/Cargo.lock
+++ b/app/src-tauri/Cargo.lock
@@ -5746,7 +5746,7 @@ dependencies = [
 
 [[package]]
 name = "ui"
-version = "0.9.0"
+version = "0.9.1"
 dependencies = [
  "arboard",
  "base64 0.22.1",
--- a/app/src-tauri/src/commands/overlay.rs
+++ b/app/src-tauri/src/commands/overlay.rs
@@ -215,6 +215,24 @@ pub fn set_overlay_expanded(app: tauri::AppHandle, state: tauri::State<'_, State
     }
 }
 
+/// Show and focus the main app window.
+///
+/// The overlay uses this instead of frontend window APIs so it does not need
+/// broad `core:window:allow-show` / `allow-set-focus` permissions.
+#[tauri::command]
+pub fn show_main_window(app: tauri::AppHandle) -> Result<(), String> {
+    match app.get_webview_window("main") {
+        Some(window) => {
+            window.show().map_err(|e| e.to_string())?;
+            window.set_focus().map_err(|e| e.to_string())
+        }
+        None => {
+            tracing::warn!(target: "system", "show_main_window: main window not found");
+            Ok(())
+        }
+    }
+}
+
 /// Hide the always-on-top overlay window.
 #[tauri::command]
 pub fn hide_overlay(app: tauri::AppHandle) -> Result<(), String> {
--- a/app/src-tauri/src/commands/overlay.rs
+++ b/app/src-tauri/src/commands/overlay.rs
@@ -7,6 +7,10 @@ use tauri::Manager;
 const NOTCH_EXPAND: f64 = 120.0; // 60px expansion room on each side
 #[cfg(target_os = "macos")]
 const FALLBACK_OVERLAY_W: f64 = 200.0;
+#[cfg(target_os = "macos")]
+const FALLBACK_OVERLAY_H: f64 = 37.0;
+#[cfg(target_os = "macos")]
+const EXPANDED_DROP: f64 = 56.0; // extra height for the hover dropdown row
 
 #[derive(serde::Serialize, Clone)]
 pub(crate) struct NotchInfo {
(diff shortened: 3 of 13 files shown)


georgenijo:
Updated this PR with three focused follow-up commits:

- `fix: open settings from overlay via rust command` — replaces overlay frontend show/focus calls with the narrow Rust `show_main_window` command, keeping overlay capabilities tight.
- `fix: stabilize notch quick controls` — adds hover collapse/watchdog behavior, rejects double-click recording while expanded, updates quick-control icons/states, and syncs auto-paste directly to the backend.
- `ux: explain paused auto-paste state` — surfaces the auto-paste/file-saving conflict in Settings when paste is configured on but effectively paused.

Local validation:

- `cd app/src-tauri && cargo check`
- `git diff --check`

Native smoke notes: hover expand/collapse works, expanded double-click logs and no-ops, settings opens without the previous `window.show not allowed` error, and auto-paste shows amber while file output suppresses effective paste.

Issue #129, opened 2026-03-04
Streaming/chunked transcription — process audio during recording


georgenijo:
## Summary

Process audio incrementally while the user is still speaking, so that when they stop recording, most of the transcription is already done. This dramatically reduces perceived latency, especially with larger models.

## Context

Current flow: record all audio → stop → VAD → whisper processes full buffer → text. For a 45-second recording, the user waits 3-5s (base.en) or 10-20s (large-v3-turbo) after stopping.

With streaming: whisper processes chunks in the background during recording. On stop, only the final chunk needs processing — perceived wait drops to ~0.5s.

## Research

See `docs/research-streaming-transcription.md` for the feasibility analysis covering:
- whisper.cpp streaming capabilities and sliding-window approach
- whisper-rs API surface for chunked inference
- Current pipeline architecture and what needs to change
- Prior art (WhisperFlow, whisper-streaming, MacWhisper)
- Chunk size / overlap strategies
- Partial result reconciliation and deduplication

## Why this matters

- Enables larger, more accurate models without painful wait times
- Enables future LLM post-processing pipeline (stream whisper output → LLM cleans up on stop) with minimal total latency
- Matches user expectations from tools like WhisperFlow

## Open questions

- Optimal chunk size and overlap for Metal GPU
- How to display partial results (overlay? main window?)
- GPU contention if inference runs while audio capture is active
- Deduplication strategy at chunk boundaries

PR #205, opened 2026-06-19
feat: live streaming transcription preview in overlay (#129)

georgenijo:
## Summary

Adds an optional, **default-off** live preview that shows partial words in the Dynamic Island overlay **while the user is still speaking**, instead of only after they stop.

This is strictly **preview-only**. The authoritative injected/clipboard text remains the existing full-audio one-shot transcription produced on stop — it is never changed, re-ordered, or replaced by this feature.

## Design constraints honored

- **Authoritative output untouched.** The final pass in `run_transcription_pipeline` is completely unchanged. The preview pass only emits a `partial-transcript` Tauri event consumed by the overlay; it never writes the clipboard, never pastes, never feeds the history/stats path.
- **Default-off.** New `livePreviewEnabled` setting defaults to `false` everywhere (TS `DEFAULT_SETTINGS`, Rust `DictationState::default`). When off, behavior is byte-for-byte the current behavior — `stop_live_preview` even no-ops its overlay-clear emit unless a preview was actually armed.
- **Whisper backend only.** `preview::should_run_preview` gates on `backend.name() == "whisper"`, so Parakeet (and any future backend) gets a graceful no-op — nothing spawns, nothing errors.
- **Append-only plumbing.** All settings changes are additive (new field + new coercion branch + new configure key), matching the existing `cleanupEnabled` / `codeVocab*` patterns.

## How it works

1. `start_native_recording` calls `maybe_start_live_preview`, which checks the gate and spawns a throttled `tokio` task. A shared `Arc<AtomicBool>` (`live_preview_active`) in `AppState` is the cancel token.
2. Every ~1.4s the task takes a **non-destructive** `audio::snapshot_samples()` (clones the in-progress buffer, resampled to 16kHz — never touches the final capture, the `active` flag, or the capture thread), selects a trailing window (default 12s), runs a partial whisper pass on a blocking thread, and emits the cleaned text.
3. Because the loop `await`s each blocking pass before sleeping again, a slow transcribe can never pile up (**skip-if-running is implicit**).
4. `stop_native_recording` / `cancel_native_recording` flip the cancel flag and emit an empty `partial-transcript` to clear the overlay, **before** the authoritative final pass runs.
5. A `recording_id` generation guard prevents a superseded session's in-flight pass from emitting stale text on rapid start-stop.

## WhisperState concurrency decision (highest-risk area)

The transcription backend stores a single cached `WhisperState` behind `AppState::backend` (`Mutex<Box<dyn TranscriptionBackend>>`). Running a preview pass concurrently with the final pass on that **same** `WhisperState` would corrupt results (`WhisperState::full` mutates internal buffers).

**Decision: the preview pass owns its OWN dedicated `WhisperContext` + `WhisperState`** (`transcriber/preview.rs` → `PreviewTranscriber`), entirely separate from the cached backend state and never reachable through `AppState::backend`. This is the same isolation pattern VAD already uses (a fresh `WhisperVadContext` per run). Consequences:

- The preview pass and the final pass can **never** share or corrupt each other's `WhisperState` — they are different objects.
- In whisper-rs 0.15.1, `WhisperState` holds its own Arc-cloned context and is `unsafe impl Send`, so `PreviewTranscriber` moves cleanly across `spawn_blocking` boundaries.
- Trade-off: extra memory + a model load for the preview engine. Acceptable because the feature is default-off and CPU-heavy by nature. Even in the rapid start-stop window where an old and new preview task briefly overlap, each owns its own state, so the only consequence is wasted work (already bounded by the cancel flag + `recording_id` guard).

## Tests

- **Rust pure logic** (`transcriber/preview.rs`, `#[cfg(test)]`): trailing-window selection (below-min skip, shorter-than-window, trailing slice, zero-rate/zero-window fallbacks), preview-text cleaning (trim, empty drop, `[BLANK_AUDIO]`/`(silence)` marker drop, internal-bracket preservation), and the on/off + backend gate. Verified green via `rustc --test` on a /tmp copy (13/13). CI compiles+runs them under `cargo test --lib`.
- **Frontend** (`settings.test.ts`): default-off, non-boolean coercion, explicit-value preservation for `livePreviewEnabled`. `npx tsc --noEmit` clean; `npm test` 62/62 green.

## Runtime verification REQUIRED

The actual streaming behavior **cannot** be verified in CI or this headless worktree — there is no live microphone audio and no rendered overlay. A human must run a **built `.app`** (`cd app && npm run tauri build`) and confirm:

1. **Default-off is invisible.** Fresh install / untouched settings: record + stop as today. Confirm no overlay preview text ever appears and the pasted/clipboard text is exactly as before.
2. **Toggle on → partial words appear.** Settings → Transcription → enable **Live Preview** (Whisper model selected, e.g. `base.en`). Start recording and speak a long sentence. Confirm partial words appear in the Dynamic Island **while still speaking**, updating roughly every ~1.4s.
3. **Final injected text is unchanged.** Stop. Confirm the pasted/clipboard text is the full, authoritative transcription (identical to preview-off) — NOT the last partial preview string.
4. **CPU acceptable.** Watch Activity Monitor / the in-app resource monitor during a long preview-on recording. Confirm CPU is within an acceptable range and recovers after stop.
5. **No crash/hang on rapid start-stop.** Rapidly start/stop/cancel recordings (keyboard and overlay double-click) with preview on. Confirm no crash, no hung "Processing" state, no stale preview text lingering from a previous recording, and the overlay clears to idle each time.
6. **Parakeet no-op.** Switch to the Parakeet model with Live Preview still on. Confirm recording works normally and no preview appears (graceful no-op, no error).

## Summary by CodeRabbit

* **New Features**
  * Added optional live transcription preview that displays partial text in the overlay while recording (disabled by default).
  * Live preview is exclusive to the Whisper backend and can be toggled in Transcription settings.
  * Preview text is for visualization only; final injected text remains unchanged.

* **Documentation**
  * Updated overlay and transcription feature documentation to explain the new live preview functionality.

### DIFF
--- a/app/src-tauri/src/audio.rs
+++ b/app/src-tauri/src/audio.rs
@@ -324,6 +324,48 @@ pub fn stop_recording() -> Result<Vec<f32>, String> {
     }
 }
 
+/// Non-destructively snapshot the audio captured **so far** in the current
+/// recording, resampled to Whisper's 16kHz mono format.
+///
+/// Used only by the optional live-preview pass (issue #129). Unlike
+/// `stop_recording`, this does NOT take or clear the buffer, does NOT touch the
+/// `active` flag or join the capture thread, and does NOT affect the
+/// authoritative final transcription in any way — it clones the current samples
+/// under the buffer lock and returns a resampled copy. Returns `None` when no
+/// recording is in progress (no buffer present).
+pub fn snapshot_samples() -> Option<Vec<f32>> {
+    let state = RECORDING_STATE.get()?;
+    let guard = state.lock().unwrap_or_else(|poisoned| {
+        tracing::warn!(target: "audio", "snapshot_samples: recording state mutex was poisoned, recovering");
+        poisoned.into_inner()
+    });
+
+    // Only snapshot while actively recording — never after stop has begun.
+    if !guard.active.load(Ordering::Relaxed) {
+        return None;
+    }
+    let buffer = guard.shared.as_ref()?;
+    let sample_rate = guard.sample_rate;
+    let raw: Vec<f32> = {
+        let samples = buffer.lock().unwrap_or_else(|poisoned| {
+            tracing::warn!(target: "audio", "snapshot_samples: samples mutex was poisoned, recovering");
+            poisoned.into_inner()
+        });
+        samples.clone()
+    };
+    // Release the recording-state lock before the (potentially larger) resample.
+    drop(guard);
+
+    if raw.is_empty() {
+        return Some(raw);
+    }
+    if sample_rate != WHISPER_SAMPLE_RATE {
+        Some(resample(&raw, sample_rate, WHISPER_SAMPLE_RATE))
+    } else {
+        Some(raw)
+    }
+}
+
 #[allow(dead_code)]
 pub fn is_recording() -> bool {
     if let Some(state) = RECORDING_STATE.get() {
--- a/app/src-tauri/src/commands/recording.rs
+++ b/app/src-tauri/src/commands/recording.rs
@@ -1,9 +1,141 @@
 use crate::{MutexExt, State};
 use crate::state::{AppState, DictationStatus};
 use crate::transcriber;
+use crate::transcriber::preview;
 use crate::{audio, audio_decode, injector, keyboard, vad};
-use std::sync::atomic::Ordering;
+use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, Ordering};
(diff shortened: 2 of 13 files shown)


georgenijo:
Closing because this preview-only design does not satisfy #129’s core goal of reducing post-stop latency. It loads a second Whisper context and spends CPU/GPU during capture while leaving the authoritative final full-audio pass unchanged. Issue #129 remains open for a true incremental solution.

Issue #190, opened 2026-05-29
Microphone permission banner shows false-negative after dev rebuild / app move (stale TCC)


georgenijo:
## Problem

The Permissions Banner can show **Microphone: Required for recording** (red) even when the microphone actually works — recording and transcription succeed. The banner reads the live TCC status, but for adhoc/linker-signed dev builds that status goes stale and reports not-granted while the underlying capture still functions.

This is the same class of stale-TCC confusion that #176 fixed for Accessibility, now surfacing on Microphone.

## How we hit it

While smoke-testing #176 with a built `Local Dictation Dev.app` (`com.localdictation.dev`):

1. The app was **rebuilt** — adhoc/linker-signed dev builds get a new cdhash on each compile (`Signature=adhoc, TeamIdentifier=not set`).
2. The `.app` was **moved to `/Applications`** — path change invalidates the adhoc TCC grant.
3. System Settings still showed the Microphone toggle ON for `Local Dictation Dev`, and recording worked ("Hello hello hello" transcribed fine)...
4. ...but the banner stayed red because `check_microphone_permission` (`AVCaptureDevice.authorizationStatus(for: .audio)`) returned not-authorized for the new identity.

`tccutil reset Microphone com.localdictation.dev` + re-grant cleared it.

## Why it matters

- The banner's live-status read is correct in principle (it answers "can this exact binary use the mic right now?"), but the **visible toggle vs live query mismatch** is confusing — identical UX problem to #176.
- Mostly a dev-build artifact (production is Developer-ID signed with a stable identity, so grants persist). But worth confirming production is unaffected, and the dev workflow is painful.

## Possible directions (to investigate)

- Confirm production (Developer-ID signed) does **not** exhibit this; document if so.
- Consider extending the troubleshooting reset action (from #176) to also cover **Microphone** — `tccutil reset Microphone <bundle-id>` + open the Microphone pane — so the recovery flow is symmetric.
- Consider detecting the toggle-on-but-query-denied mismatch and surfacing a clearer hint.

## Acceptance criteria (proposed)

- Root cause confirmed and documented (dev-build TCC staleness vs production behavior).
- Decision on whether to add a Microphone reset action mirroring #176.
- Banner copy/behavior does not mislead when mic works but status reads stale.

## References

- #176 (Accessibility reset button) — same stale-TCC pattern, Accessibility side
- `app/src-tauri/src/commands/permissions.rs` — `check_microphone_permission`, `reset_accessibility_permission`
[...]

PR #214, opened 2026-07-16
feat: first-launch setup assistant (guided permissions + model onboarding)

georgenijo:
## Summary

Replaces the first-launch experience — previously a lone model-download screen with permissions relegated to a dismissible banner — with a guided setup assistant: **Welcome → Microphone → Accessibility → Model download → Done**.

### Highlights
- **In-app mic prompt**: new `request_microphone_access` command calls `AVCaptureDevice.requestAccess` (block2 completion handler) so the native TCC dialog fires when the user clicks Allow — not on their first recording attempt. Never opens the device, so no audio ducking (#177).
- **Live permission polling**: both permission steps poll every 1s + on focus, so a grant made during the System Settings roundtrip flips the step when the user returns.
- **Stale-TCC recovery in-flow**: denied mic gets Open Settings + a reset path; since `tccutil reset` returns the status to `notDetermined`, the in-app dialog re-arms after reset. Accessibility gets the listed-but-stale reset path (#190-family drift states).
- **Grandfathering**: existing installs with mic + accessibility + model already in place get the completion flag set silently — upgrades never see the wizard.
- **Recovery entry point**: Settings → About → **Run Setup Assistant** re-launches the wizard when a permission was revoked or drifted.
- `ModelDownloadPanel` extracted from `ModelDownloader` and embedded as the wizard's model step (Back locked during an active download to prevent concurrent `download_model` runs).

### Not changed
- `PermissionsBanner` remains as the post-onboarding drift catcher.
- Standalone `ModelDownloader` gate remains for a deleted model file after onboarding.

## Test plan
- [x] `npx tsc --noEmit` clean (Linux)
- [x] `npm test` 70/70 (Linux)
- [x] `cargo check` clean on macOS (Mac Mini)
- [x] `cargo test --lib -- --test-threads=1` 249/249 on macOS (Mac Mini)
- [ ] Manual: fresh-TCC run on a real Mac (`tccutil reset Microphone/Accessibility` + `build`) to see the dialogs end-to-end

## Summary by CodeRabbit

* **New Features**
  * Added a first-launch setup assistant guiding users through microphone access, accessibility permissions, and model selection or download.
  * Added inline recovery options for denied or outdated permission states.
  * Existing installations that are already configured continue without interruption.
  * Added **Run Setup Assistant** under Settings → About to restart setup at any time.

* **Bug Fixes**
  * Recording controls remain unavailable until required setup steps are complete.
  * Permission status updates automatically when returning from System Settings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Added
+- **First-launch setup assistant** — new installs get a guided wizard (Welcome → Microphone → Accessibility → Model download → Done) instead of a dismissible permissions banner next to a lone model-download screen. The microphone step fires the native macOS permission dialog in-app (new `request_microphone_access` command via `AVCaptureDevice.requestAccess`) instead of waiting for the first recording attempt; both permission steps poll live so a grant made in System Settings flips the step when you come back, and denied/stale-TCC states get inline reset-and-retry paths. Existing installs with permissions and a model already in place are grandfathered silently. Re-run anytime via Settings → About → Run Setup Assistant (`OnboardingFlow.tsx`, `lib/onboarding.ts`).
+
 ## [0.14.1] - 2026-07-16
 
 ### Changed
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -16,6 +16,7 @@ cd app && npx tsc --noEmit         # TypeScript check
 Read these before working on a feature:
 
 - **[docs/onboarding.md](docs/onboarding.md)** — Setup, permissions, model installation, logs
+- **[docs/features/onboarding-flow.md](docs/features/onboarding-flow.md)** — First-launch setup assistant (permissions wizard + model download)
 - **[docs/features/recording-modes.md](docs/features/recording-modes.md)** — Hold-down and double-tap modes, state machine, rdev threading
 - **[docs/features/transcription.md](docs/features/transcription.md)** — Audio capture, whisper pipeline, status flow
 - **[docs/features/text-injection.md](docs/features/text-injection.md)** — Clipboard, auto-paste, osascript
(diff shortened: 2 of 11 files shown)


georgenijo:
@cursor-agent review

georgenijo:
@cursoragent review — post findings only, don't push fixes

georgenijo:
Addressed in 5a73ba9 (@cursoragent findings):

1. ✅ **Hotkeys during wizard** — all three toggle hooks + Escape cancel now also gate on `onboardingState === 'done'` (`hotkeysArmed`).
2. ✅ **Stuck `onDownloadingChange(true)`** — `listen()` + `download_model` now share one try/catch; the flag and the unlisten ref are always cleared on any failure.
3. ✅ **Hardcoded 'Left Shift'** — `OnboardingFlow` now receives `recordingMode` + `triggerKey` and renders the real binding (hold / double-tap / both copy variants).
4. ✅ **Stale poll responses** — monotonic sequence counter; out-of-order probe results are dropped.
5. ✅ **Model summary row** — binds to `modelInstalled === true` with an honest fallback message.
6. ⏭️ **Fail-open model probe** — keeping as-is, deliberately: it mirrors the standalone `ModelDownloader` gate's fail-open (`App.tsx` mount check), and `check_specific_model_exists` is a local file-existence check that only throws if the IPC layer itself is broken — in which case the download flow wouldn't work either. If the model is genuinely missing the probe resolves `false` and the wizard downloads; lazy model load still surfaces a clear error at first transcription in the worst case.

georgenijo:
@coderabbitai review

georgenijo:
Addressed the CodeRabbit nitpicks in the latest commit: added typed wrappers to `lib/dictation.ts` (`checkAccessibilityPermission`, `requestAccessibilityPermission`, `requestMicrophoneAccess`, `openMicrophoneSettings`, `checkModelExists`) and removed all direct `invoke()` usage from `OnboardingFlow.tsx` and the onboarding/model gates in `App.tsx`. The remaining direct `invoke` sites CodeRabbit flagged in `PermissionsBanner.tsx` / `SettingsPanel.tsx` / `ModelDownloader.tsx` (`download_model`) predate this PR — leaving those for a follow-up sweep to keep this diff scoped.

PR #218, opened 2026-07-17
fix: recover interrupted model installations

### LINKED ISSUE #215: Parakeet v2 install is lost when extraction is interrupted
## Bug

Performance Lab returns Parakeet v2 to an uninstalled Download state after its 1.2 GB download completes if Murmur exits while unpacking the archive.

## Evidence

- The retained archive is exactly 1,120,982,957 bytes, matching the server Content-Length.
- `bzip2 -t` validates the archive successfully.
- The final model directory is partial: encoder and tokens exist, but decoder and joiner are missing.
- On restart, the completeness check correctly returns false, but retry ignores the complete archive and would download 1.2 GB again.

## Expected

Installation should extract into a staging directory, publish the model atomically only after validation, retain/reuse a complete downloaded archive across an interrupted extraction, and show an installing phase after network progress reaches 100%.

## Reported in

v0.16.0 Performance Lab.

### LINKED ISSUE #217: Core ML model setup shows a frozen 0% progress bar
## Bug

The first-launch/onboarding model step initializes FluidAudio Core ML setup with `{ received: 0, total: 0 }`. The shared downloader renders that unknown total as a zero-width determinate bar, so Parakeet v3 appears frozen at 0% until setup completes and the wizard abruptly advances. Settings and Performance Lab have the same unknown-progress behavior.

## Constraint

FluidAudio currently exposes only a synchronous `init_asr()` call and no byte-progress callback, so a numeric percentage would be fabricated.

## Expected

All model download surfaces should render unknown-total Core ML setup as a clearly animated indeterminate Installing state, while Whisper/sherpa downloads retain honest byte percentages.

georgenijo:
## Summary
- recover complete Parakeet v2 archives left by v0.16.0 instead of downloading them again
- extract and validate model bundles in staging before atomically publishing them
- serialize same-model installs and the shared VAD transaction
- distinguish invalid archives from transient install failures so retries reuse good local data
- show indeterminate Installing progress for Core ML setup across onboarding, Settings, and Performance Lab
- reject truncated streamed downloads

## Verification
- cargo test -- --test-threads=1 (261 passed; Core ML hardware test ignored)
- npm test -- --run (78 passed)
- npm run build
- npx tsc --noEmit
- bzip2 -tv on the retained 1.12 GB v2 archive

Closes #215
Closes #217

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Fixed
+- Parakeet v2 downloads now survive an interrupted extraction: Murmur reuses the completed archive, validates a staged bundle, and publishes it atomically instead of leaving a partial model that appears undownloaded (#215).
+- Core ML model setup now shows an animated indeterminate Installing state across onboarding, Settings, and Performance Lab instead of a frozen 0% bar (#217).
+
 ## [0.16.0] - 2026-07-17
 
 ### Added
--- a/app/src-tauri/src/commands/models.rs
+++ b/app/src-tauri/src/commands/models.rs
@@ -1,7 +1,22 @@
 use crate::{MutexExt, State};
 use crate::transcriber::{self, TranscriptionBackend};
 use crate::vad;
+use std::collections::HashMap;
+use std::sync::{Arc, LazyLock, Mutex};
 use tauri::Emitter;
+
+type ModelInstallLock = Arc<tokio::sync::Mutex<()>>;
+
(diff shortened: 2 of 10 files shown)


georgenijo:
Addressed all four CodeRabbit findings in 73cd69e:

- same-model installs are single-flight; VAD has its own shared lock
- typed archive errors now redownload corrupt content but preserve archives on transient install/publish failures, with tests for both paths
- reduced-motion uses a static full-width indicator with no animation
- the Download Pipeline docs now describe staging, validation, atomic publication, and retention correctly

Full local verification: 261 Rust tests, Whisper integration, 78 frontend tests, TypeScript, and production frontend build.

Issue #270, opened 2026-07-20
Benchmark WER punishes text formatting, not recognition — normalize before scoring


georgenijo:
## Problem

`word_errors()` in `app/src-tauri/src/benchmark.rs:353` lowercases and splits on non-alphanumeric, but has no text normalization. "16 kHz" vs reference "sixteen kilohertz" counts as errors; "Mac OS" vs "macOS", "front end" vs "frontend" likewise. These are formatting/ITN differences, not misrecognitions.

## Proof (from the 2026-07-19 v0.17.2 production runs)

Scorer reimplemented and validated: all 28 per-fixture error counts reproduce exactly. Adding standard normalization (digits→words, unit abbreviations, compound joins — Whisper EnglishTextNormalizer-style):

| model | raw WER (shipped) | normalized |
|---|---|---|
| Parakeet v3 | 8.4% | **4.5%** |
| Parakeet v2 | 3.9% | 1.3% |
| tiny.en | 14.3% | 10.4% |
| base.en | 16.2% | 8.4% |
| small.en | 5.2% | **1.3%** |
| medium.en | 3.9% | 1.3% |
| large-v3-turbo | 2.6% | 1.3% |

Consequences of the current raw metric:
- Half of Parakeet v3's reported errors are formatting artifacts (its digit/unit style differs from the fixture reference style).
- small.en is actually tied with v2/medium/turbo, not 2x worse — the current metric hides the best value model in the whisper family.
- Downstream, the `mostAccurate` and `balanced` recommendations rank on distorted numbers.

## Proposed fix

Normalize both reference and hypothesis before the Levenshtein pass in `word_errors()`: digit↔word equivalence, common unit abbreviations, compound-word joins. Keep the raw transcript display unchanged so users still see actual output.

🤖 Filed from a Claude Code investigation session

PR #289, opened 2026-07-20
fix: stabilize Performance Lab recommendations

### LINKED ISSUE #272: 'fastest' recommendation flips between identical runs — pooled-median artifact
## Problem

Two production benchmark runs on the same machine, same build (v0.17.2, 2026-07-19, 06:47 and 14:09) produced different `fastest` picks: `tiny.en` then `parakeet-tdt-0.6b-v3-coreml`. Nothing changed — the metric is noise-sensitive.

`recommendations()` in `benchmark.rs:468` ranks by the model-level `warm_median_ms`, which (`benchmark.rs:680`) is a percentile over ALL warm samples pooled across fixtures — 2s and 28s audio dumped into one distribution. The median lands between fixture strata and jitters across runs (v3: 108.6ms vs 77.5ms pooled median between runs).

## Proof

`realtimeFactor` — already computed per model in the same reports — picks Parakeet v3 in BOTH runs:

| run | pooled-median winner | realtimeFactor winner |
|---|---|---|
| 06:47 | tiny.en (79.1 vs v3 108.6) | **v3** (0.0080 vs tiny 0.0086) |
| 14:09 | v3 (77.5 vs tiny 82.4) | **v3** (0.0082 vs tiny 0.0092) |

## Proposed fixes

1. `fastest` = min `realtime_factor` (one-line change; metric is speed-per-audio-second, comparable across fixture mixes).
2. `balanced` (`benchmark.rs:482`) never considers memory → currently picks Parakeet v2 at **2,925 MB** RSS delta over alternatives at 46–545 MB. Add a memory tie-break or cost term within the accuracy window.
3. Optional: treat sub-10% latency deltas as ties instead of presenting noise as signal.

🤖 Filed from a Claude Code investigation session

georgenijo:
## Summary
- rank Fastest by duration-weighted realtime factor instead of pooled warm median
- keep Balanced within the two-point WER window, treat speeds within 10% as equivalent, then prefer lower memory with deterministic tie-breaks
- document the recommendation policy in Performance Lab UI/docs and cover flip, memory, boundary, incomplete-result, and deterministic-tie regressions

## Test plan
- [x] `cargo check`
- [x] `cargo test -- --test-threads=1` (371 unit tests and available Whisper integration test passed; optional Core ML integration remained model/env gated)
- [x] `npx tsc --noEmit`
- [x] `npm test -- --run` (19 files, 136 tests)
- [x] `git diff --check`
- [x] Debug native bundle produced; expected signing-key error occurred only after the .app, DMG, and updater archive were built
- [x] Native Performance Lab Quick run on final v0.18.0 build with Parakeet v3 + Whisper Base: 76x vs 47x, both 0.0% WER, Fastest/Accurate/Balanced = Parakeet v3, updated explanation rendered

## Parallel PR note
PR #282 is a parallel implementation. It applies the 10% memory-aware tie band to Fastest itself and changes only Rust. This PR keeps Fastest as the strict minimum realtime factor from issue fix #1, applies the speed-equivalence/memory policy only to Balanced, and updates the visible contract and feature documentation as approved.

Closes #272


georgenijo:
Heads-up: #294 (the benchmark trust wave) just merged and includes an equivalent fix for #272 — same core approach as this PR (rank by realtime factor, ~10% tie band, memory tie-break), plus one thing this branch can't have: the `balanced` accuracy window now filters on the **normalized** WER introduced by #270, which didn't exist on main when this was authored. The shared tie-break helper lives in `pick_by_metric_with_tiebreak()` (benchmark.rs). Issue #272 is closed by that merge, and this branch will now conflict — suggest closing this PR as superseded, but flagging rather than closing since it's yours. If anything here is stricter than what landed (e.g. the duration-weighted corpus speed wording), happy to fold it in as a follow-up.

🤖 From the benchmark orchestration session

PR #343, opened 2026-07-22
fix(transform): clipboard-fallback hardening — fail-closed secure check, full pasteboard restore, changeCount guard, honest errors (#334 #335 #336)

### LINKED ISSUE #334: Transform: clipboard fallback breaks secure-field fail-closed — fires after an errored AXSubrole check
## Summary

The clipboard-fallback capture path is entered when the secure-field AX check **itself** errors or times out, violating the documented "AX errors during the check fail closed" invariant. The check fails **over** to a synthetic Cmd+C rather than failing closed.

**Severity:** Medium (security invariant / defense-in-depth regression). Not a mainstream password leak — see scope.
**Found in:** review of #325/#327/#329, branch `issue/329-mic-prearm` @ `db5d99e`.

## Behavior

`capture_selection` classifies a non-benign AXSubrole/AXRole status (incl. the 25ms timeout `-25204`, `is_benign_role_query_error` at `selection.rs:184-186`) as `AxUnavailable` (`selection.rs:750-767`). The match arm at `selection.rs:301-315` then routes exactly `NoSelection | AxUnavailable` into `clipboard_fallback::capture_via_clipboard`, which has **zero** secure-field awareness — it posts a global synthetic Cmd+C (`:385`). The secure check reads the AX **focused element** (`:741`); Cmd+C is serviced by the app's **copy responder**, which need not be that element.

Only `SecureField` and `AccessibilityDenied` remain terminal (`:265-279`, `:316`).

## Repro (deterministic to observe the fail-over)

1. Focus a password field in an AX-slow app (browser password input).
2. Hold right Option, speak, release.
3. Logs show `AX capture incomplete — attempting clipboard fallback` with an `ax_outcome`, on a secure field — instead of a hard secure-field refusal.

## Scope / why not High

Standard secure widgets (NSSecureTextField, Chromium/WebKit password inputs, Qt `QLineEdit` password echo, Swing `JPasswordField`) suppress Copy at framework level, so a synthetic Cmd+C returns nothing → sentinel unchanged → timeout → no capture. Leakage requires a **custom-drawn / masked** secure field that does not suppress Copy, in an AX-slow app. Recipient is the local popover + local sidecar, not cloud. So: a real invariant break and defense-in-depth regression, low real-world exploitability.

Doc drift: `docs/features/selected-text-transform.md:55` ("AX errors during the check fail closed") is now false and should be reconciled with the fallback described at line 67.

## Suggested fix

Distinguish "secure check itself failed" from "focused-element query failed" and fail **closed** for the former (do not fall back on an errored secure-subrole read). At minimum, re-run the secure-subrole check on the copy-responder element before accepting fallback text.

---

### LINKED ISSUE #335: Transform: clipboard fallback destroys non-text clipboard and ingests foreign clipboard writes as the selection
## Summary

The clipboard-fallback capture path mishandles the user's pasteboard in two ways: it **destroys non-text clipboard contents**, and it **ingests foreign clipboard writes** as the "selection" because the poll has no tie between the text it reads and the Cmd+C it posted.

**Severity:** Medium (user data loss + wrong-content ingestion into the LLM).
**Found in:** review of #325/#327/#329, branch `issue/329-mic-prearm` @ `db5d99e`.

## Defect A — non-text clipboard destroyed

`selection.rs:421` snapshots via `read_clipboard_text().ok()` (arboard text-only, `injector.rs:4-10`) → `None` for an image/file clipboard. The sentinel write at `:423` then overwrites it unconditionally; the restore match (`:440-449`) with `None` leaves either the captured selection text or `""`. Every fallback run with a non-text clipboard loses it.

### Repro (deterministic)
1. Screenshot to clipboard: Cmd-Ctrl-Shift-4, grab a region.
2. In Brave, select text; run a transform (hold right Option, speak, release) — browser forces the fallback.
3. After it completes, Cmd-V into Notes → the image is gone; you paste selection text or nothing.

## Defect B — foreign clipboard writer ingested

`poll_for_selection` (`selection.rs:394-415`) accepts any text where `text != sentinel` (`:406-408`). No NSPasteboard `changeCount`, no owner/source check ties the read to the synthetic Cmd+C. A clipboard manager reacting to the sentinel write, or a Universal Clipboard push from iPhone during the 300ms window, becomes `TransformSnapshot.text` (`:451-469`) → popover + sidecar.

### Repro (deterministic with helper)
```bash
while true; do printf 'INJECTED-%s' $RANDOM | pbcopy; sleep 0.1; done
```
Run that on the Mac, then select text in Brave and do a fallback capture. The popover proposal is built from `INJECTED-####`, text you never selected. Kill the loop after.

## Suggested fix

- Snapshot and restore the **full** pasteboard (all types via NSPasteboard item list), not text-only, so images/files survive.
- Read NSPasteboard `changeCount` before posting Cmd+C and require the accepted read to be the change produced by that copy (changeCount incremented exactly once past the sentinel write), rejecting concurrent third-party writes.

---

### LINKED ISSUE #336: Transform: all clipboard-fallback failures report 'Select some text first' even when text is selected
## Summary

Every clipboard-fallback failure collapses to `SelectionError::NoSelection`, so the popover tells the user "Select some text first" even when text **is** clearly selected. This is the most frequently-hit path in practice (browser captures are ~24/27 fallback).

**Severity:** Medium (misleading UX, recurring).
**Found in:** review of #325/#327/#329, branch `issue/329-mic-prearm` @ `db5d99e`.

## Repro (deterministic)

1. In Brave, select clearly-visible text.
2. Hold right Option, speak, release.
3. Popover shows **"Select some text first"** despite the visible selection.

## Root cause

`selection.rs:451` returns `copied.ok_or(SelectionError::NoSelection)`; a sentinel-write failure also maps to `NoSelection` (`:423-424`). The original `AxUnavailable` survives only if the `spawn_blocking` join itself fails (`:314`, `unwrap_or(Err(err))`) — a near-dead branch. In Chromium-family apps the synthetic Cmd+C is swallowed because those apps read live hardware modifier state and see Cmd+Opt+C while the user holds right Option (the module documents this at `selection.rs:333-340`), so the poll times out and the user is told to select text they already selected.

Error surfaces via `selection_error_code` (`transform_flow.rs:98-104`) → popover copy `transformReview.ts:66` (`no_selection: 'Select some text first'`) vs the more accurate `:68` (`ax_unavailable: "Couldn't read the selection"`), which is effectively unreachable.

## Suggested fix

Preserve the real terminal error: when the AX path ended in `AxUnavailable` and the fallback then failed, return `AxUnavailable` (not `NoSelection`) so the popover shows "Couldn't read the selection." Reserve `NoSelection` for the genuinely-empty case.

---

georgenijo:
## Summary

Hardens the transform clipboard-fallback capture path (`selection.rs`), fixing three issues from the multi-agent transform review:

### #334 — secure-field fail-closed restored
An errored secure-field check (non-benign AXSubrole/AXRole status incl. the 25ms messaging timeout, or a failed messaging-timeout arm on the focused element) is now a distinct terminal outcome, `SelectionError::SecureCheckFailed`:
- **retry-eligible** — the AX warm-up retry loop may retry it, since a retry re-runs the full secure check and reads nothing unless it passes (this keeps Chromium's lazy-AX warm-up path working);
- **never fallback-eligible** — no synthetic Cmd+C at an element we could not prove is not a password field. Pure `fallback_eligible()` / `retry_eligible()` helpers now gate both decisions and are unit-pinned.
- Popover shows the accurate `ax_unavailable` copy ("Couldn't read the selection"); logs carry the distinct `secure_check_failed` outcome. Feature doc reconciled.

### #335 — pasteboard fidelity + foreign-write rejection
- **Defect A:** the fallback now snapshots and restores the **full pasteboard** (every item, every type, via `NSPasteboard`/`NSPasteboardItem`) instead of text-only — screenshots/files survive the sentinel dance. Restore happens on success *and* failure.
- **Defect B:** the poll now records `changeCount` after the sentinel write and accepts a copy only when the count moved **exactly one step**, re-checked after the text read. A clipboard manager or Universal Clipboard push interleaving in the 300ms window is rejected instead of being ingested into the LLM as the "selection".

### #336 — honest failure errors
When the fallback itself fails, the AX path's terminal error is preserved (`fallback_failure_error`): browser users with a visible selection now see "Couldn't read the selection" (`ax_unavailable`) instead of the misleading "Select some text first". `no_selection` is reserved for the genuinely-empty case.

## Regression proof (each mutation reproduces the pre-fix behavior; run on this branch)

| Pre-fix behavior restored by mutation | Result |
|---|---|
| #334: `SecureCheckFailed` allowed to fall back | `secure_check_failure_is_never_fallback_eligible` **FAILED** |
| #335A: text-only pasteboard restore | `pasteboard_snapshot_restores_non_text_content_in_full` **FAILED** (real-NSPasteboard round-trip w/ custom binary type) |
| #335B: accept any non-sentinel text, no changeCount guard | 2 tests **FAILED** (`…rejects_foreign_writer_that_bumps_past_our_copy`, `…rejects_foreign_write_racing_the_text_read`) |
| #336: collapse failures to `NoSelection` | `fallback_failure_preserves_the_ax_paths_terminal_error` **FAILED** |
| Unmutated | 22/22 selection tests pass |

## Verification

- `cargo test -- --test-threads=1` — 557 passed (7 new tests)
- `npx tsc --noEmit` — clean
- `npx vitest run` — 272 passed
- App crate still never links llama; no network; logging stays content-free (`secure_check_failed` outcome string only, no content fields added).

Closes #334
Closes #335
Closes #336


georgenijo:
Adversarial (fable) review of the first commit confirmed 3 defects; fixed in the follow-up commit:

- **D1 (fixed):** fallback eligibility judged only the final retry attempt's error — an errored secure check on attempt 1 could be laundered back into fallback eligibility by a later, shallower failure. The secure-check error is now **latched across the whole retry ladder** (`fallback_allowed`), unit-pinned by `secure_check_error_bars_fallback_for_the_whole_capture` (mutation-verified: removing the latch fails the test).
- **D3 (fixed):** arboard clears the pasteboard *before* writing, so a failed sentinel write could still have destroyed the contents without restore. That path now restores the snapshot too, and a refused `writeObjects` restore logs content-free (item count only).
- **D2 (documented residual):** `changeCount` cannot attribute a writer — a **single** quiet foreign write inside the 300ms window while the Cmd+C is swallowed remains indistinguishable from a real copy. No NSPasteboard API closes this fully; now stated honestly in the module comment and the feature doc (the guard reduces exposure from "any write in the window" to that single-quiet-write case).

Suite: cargo 558 passed / tsc clean / vitest 272 passed.

Issue #340, opened 2026-07-22
Transform: reduce end-to-end latency (warm-on-arm, skip AX retry on browsers, token streaming)


georgenijo:
## Summary

Transform end-to-end latency is dominated by avoidable costs, not by the LLM. The sidecar is already Metal-offloaded (`with_n_gpu_layers(1_000)`, `sidecars/local-llm/src/main.rs:233`) and greedy — steady-state generation (~9s / ~50 tokens for Qwen2.5-1.5B-Q4) is near the model's ceiling. The wins are around it.

**Type:** Enhancement.
**Found in:** review of #325/#327/#329, branch `issue/329-mic-prearm` @ `db5d99e`. Observed on a live install: first-use ~34s, then ~17s, then steady ~9s; browser captures add ~2s before the LLM starts.

## Cost breakdown & levers

**1. Cold-load cliff (biggest, fully avoidable).** `idle_after = 5 min` (`llm_sidecar.rs:449`), `handshake_timeout = 30s` (`:443`). After 5 min idle the helper unloads and the next transform pays full model **load** — that is the 34s→17s→9s curve (load, not generation). Fix: **warm-on-arm** — the mic is already pre-armed on transform-key-down (`transform_flow.rs:855-874`); spawn + preload the sidecar in that same window so the model is resident by the time instruction-ASR completes. Alternatives: keep warm while the transform listener is active, or lengthen `idle_after`. Trades RSS/battery, so warm-on-hold beats always-on.

**2. ~2s wasted per browser capture.** The AX retry ladder (2×250ms gaps + 3 attempts, `selection.rs:242-280`) runs before the fallback on every browser press, and browsers hit the fallback ~24/27 of the time anyway. Detect a known-Chromium bundle id and skip straight to the clipboard fallback. Cuts ~1.5-2s off the front of every browser transform.

**3. Perceived latency.** Stream partial tokens from the sidecar to the popover instead of spinner-then-pop (protocol change: emit tokens as generated). No wall-clock change, large felt-latency change.

**4. Free win from the #1 bug fix.** Releasing `recording_transition` before the sidecar await (the fix for the review-destruction bug) also stops the whole pipeline serializing behind one up-to-20s call.

Raw generation speed is already near model-max on Metal; only a smaller/draft model or speculative decoding would move it, which is a larger scope and probably not worth it.

## Suggested order

1. Warm-on-arm (kills the cold cliff — the dominant cost).
[...]

Issue #327, opened 2026-07-21
Failed transform review popover blocks dictation until manually dismissed


georgenijo:
## Symptom

After a transform pass fails (sidecar crash, no instruction, capture error, model missing), the review popover shows the error — and until the user explicitly dismisses it (Esc / press the transform key again), pressing the dictation key does nothing:

```
WARN pipeline: start_native_recording: blocked — transform in progress
```

Found during #325 native smoke testing on a built .app.

## Root cause

A failed review parks the flow at `TransformStatus::ReviewPending` (failure is only `ReviewState::Failed` + no proposed text in the session — there is no `Failed` status variant). `blocks_recording()` is `status != Idle`, so the failed popover pins dictation even though nothing user-approvable is at stake.

## Fix

On `start_native_recording`, if the transform flow is parked on a **failed** review — `ReviewPending` with no proposed text — auto-dismiss it (atomic `try_transition(ReviewPending → Idle)`, clear session, hide popover) and let the recording start. A **ready** review (proposal present) keeps blocking: it may hold real work the user hasn't accepted, and dismissing it implicitly would destroy it.

`blocks_recording()` itself stays unchanged, so the existing `every_non_idle_transform_status_blocks_recording_start` invariant holds — the fix clears the status first when it is provably safe.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #345, opened 2026-07-22
fix(transform): dictation keypress during 'thinking' no longer destroys the ready review (#333)

### LINKED ISSUE #333: Transform: dictation keypress during 'thinking' silently destroys the ready review (recording_transition held across sidecar await)
## Summary

A dictation keypress made while a transform is still "thinking" silently destroys the ready review the user was waiting for — often many seconds after the physical keypress. Confirmed by hand (the review timed out / vanished on approve).

**Severity:** High (silent destruction of user work + latency root cause).
**Found in:** review of #325/#327/#329, branch `issue/329-mic-prearm` @ `db5d99e`. Line numbers at that commit.

## Repro (deterministic)

1. Select a **long** paragraph in Notes (longer selection → slower sidecar → bigger window).
2. Hold **right Option**, say "make this shorter", release. Popover shows **thinking**.
3. While it still says thinking, **double-tap left Shift** (dictation).
4. When the sidecar finishes, the ready proposal appears and is instantly torn down; a dictation recording starts (or a ~0-length "no speech" flash). You never get to approve.

Pre-#327/#329 the same press returned `busy_transforming` and the review survived.

## Root cause

`finish_transform_instruction` acquires the FIFO-fair `recording_transition` tokio mutex (`transform_flow.rs:920`) and holds it across whisper transcription (`:947`) **and** the sidecar await inside `run_transform` (`:951-960`, `join.await` at `:575-581`, `DEFAULT_TRANSFORM_DEADLINE = 20s` at `:64`).

`start_native_recording` parks on that same mutex at its very first statement (`recording.rs:1930`), **before** the "active transform blocks recording" guard (`:1979`) and before the new dismiss block (`:1947`). So a press during Thinking queues; when the transform releases the lock the queued press resumes, observes the freshly-set `ReviewPending`, and `dismiss_review_for_recording` (`transform_flow.rs:1165-1178`) destroys it. No freshness check exists — the press carries only `device_name` (`recording.rs:1922-1926`), no timestamp. In hold-down mode the matching release also queues (`stop_native_recording`, `:2074`), adding a stray ~0-length recording.

The lock-hold is pre-existing; the destructive dismiss-on-resume is the in-range change (ef9687c failed-only, c766911 extended to ready).

## Suggested fix

- Do not hold `recording_transition` across the sidecar await. Take it only for the state transition, release before the LLM call, and gate concurrent starts on `TransformStatus` instead (a press arriving during Thinking then hits the `blocks_recording()` guard and is refused, as before).
- Or: timestamp the dictation press and refuse dismissal of any review that became `ready` after the press was issued.

Preferred is the first — it also removes the full-pipeline serialization behind one 20s call (see the latency issue).

---

georgenijo:
## Summary

Fixes #333: `finish_transform_instruction` held the FIFO-fair `recording_transition` tokio mutex across the whisper instruction-transcription **and** the sidecar await (up to `DEFAULT_TRANSFORM_DEADLINE` = 20s). `start_native_recording` parks on that mutex at its very first statement — before the "active transform blocks recording" guard — so a dictation keypress made while a transform was still *thinking* queued, resumed the moment the transform released the lock, observed the freshly-set `ReviewPending`, and destroyed the ready review via `dismiss_review_for_recording`. Silently, seconds after the physical press. In hold-down mode the matching release also queued, adding a stray ~0-length recording.

**Fix (the issue's preferred option):** the lock is now scoped to the audio-stop + `Listening → Thinking` transition only and released before the heavy awaits. A press during Thinking acquires the mutex immediately, hits the `blocks_recording()` guard, and is refused with `busy_transforming` — the pre-#327 behavior — so the review survives. Concurrency during Thinking is gated by `TransformStatus`, which every work-starting entry point checks. Ordering inside the lock is unchanged (status check → stop audio → enter_thinking), so the mic-leak and cancel-race semantics are untouched. This also removes the full-pipeline serialization behind one 20s call (#340 item 4).

## Tier: needs-live-smoke — DO NOT MERGE until a human runs the checklist

The defect is a cross-task lock-ordering race with real rdev/audio/sidecar timing; it has no headless regression test (the command layer is pinned to the Wry runtime). Unit suites only prove no regression: `cargo test -- --test-threads=1` 550 passed, `tsc` clean, `vitest` 272 passed.

### Live smoke checklist (built .app, real model)

- [ ] **The #333 repro, fixed:** select a long paragraph in Notes → hold right Option, say "make this shorter", release → while the popover still says **thinking**, double-tap left Shift (dictation). Expected: overlay flashes the transform-busy refusal immediately; when the sidecar finishes, the **ready review appears and stays**; Approve works. No dictation recording starts, no ~0-length "no speech" flash.
- [ ] Same during **listening** (hold, speak, don't release; double-tap Shift with the other hand): dictation refused, transform continues.
- [ ] Press dictation with a **ready** review parked (no transform in flight): review is auto-dismissed and recording starts (the #327/#329 behavior, unchanged).
- [ ] Esc / short-tap cancel mid-thinking still cancels promptly; a dictation press right after cancel starts recording cleanly.
- [ ] Approve → paste-fallback app (Brave): apply works; Undo restores; clipboard restored.
- [ ] Hold-down dictation mode: press-and-hold during thinking → on release, no stray recording artifacts.

Closes #333

### DIFF
--- a/app/src-tauri/src/transform_flow.rs
+++ b/app/src-tauri/src/transform_flow.rs
@@ -922,23 +922,40 @@ pub(crate) async fn finish_transform_instruction(
     app_handle: tauri::AppHandle,
     state: tauri::State<'_, crate::State>,
 ) -> Result<(), String> {
-    let _transition = state.app_state.recording_transition.lock().await;
-
-    // Tolerate a stray release: only proceed from Listening.
-    if state.app_state.transform_status() != TransformStatus::Listening {
-        return Ok(());
-    }
-
-    let samples = crate::audio::stop_recording().unwrap_or_default();
-
     let fx = TauriFlowEffects {
         app: &app_handle,
         state: &state,
     };
-    if !enter_thinking(&state.app_state, &fx) {
-        // Lost the race to a concurrent cancel.
-        return Ok(());
-    }
+
+    // Hold `recording_transition` ONLY for the audio-stop + state transition,
+    // never across the whisper transcription or the sidecar await (issue
+    // #333). The FIFO-fair mutex means anything parked on it resumes the
+    // moment we release; pre-fix, a dictation keypress made during a long
+    // Thinking phase queued here for up to ~20s+, resumed after the review
+    // reached `ready`, and destroyed it via the parked-review auto-dismiss —
+    // silently, seconds after the physical press. With the lock released
+    // before the heavy work, that press acquires the mutex immediately,
+    // observes `Thinking` via the `blocks_recording()` guard, and is refused
+    // with `busy_transforming` — the pre-#327 behavior — so the review
+    // survives. Concurrency during Thinking is gated by `TransformStatus`,
+    // not by this mutex: every work-starting entry point refuses while the
+    // status is non-Idle.
+    let samples = {
+        let _transition = state.app_state.recording_transition.lock().await;
+
+        // Tolerate a stray release: only proceed from Listening.
+        if state.app_state.transform_status() != TransformStatus::Listening {
+            return Ok(());
+        }
+
+        let samples = crate::audio::stop_recording().unwrap_or_default();
+
+        if !enter_thinking(&state.app_state, &fx) {
+            // Lost the race to a concurrent cancel.
+            return Ok(());
+        }
+        samples
+    };
 
     let original = match transform_apply::session_snapshot(&state.app_state) {
         Some(session) => session.snapshot.text,


georgenijo:
Rebased onto main (now includes #331 mic-prearm, #343, #344); full suite green on the combined code (cargo 553 / tsc / vitest 272). Ready for the smoke checklist above.

georgenijo:
Live smoke (George, built .app, release sidecar): **key race test PASSED** — long-paragraph transform in Notes, double-tap left Shift during *thinking* → dictation press rejected, review appeared and stayed, Approve/cleanup worked. Multiple supersede passes and Esc spam during thinking produced no stuck state or stray recordings. (Esc-during-thinking not cancelling turned out to be a pre-existing gap on main, filed as #347 — not this PR.) Merging on George's authorization.

Issue #329, opened 2026-07-21
Transform UX map: browser selection fallback, silent hotkey rejection, main-window activation leak


georgenijo:
Findings from #325/#327 native smoke on macOS 26 (built .app, logs content-free throughout). The transform pipeline itself works end-to-end in AX-native apps (TextEdit: capture → instruction → sidecar → ready diff → approve). The remaining friction is UX-shaped:

## 1. Browser selection capture fails (Chromium/Brave, YouTube pages)
`Couldn't read the selection` — Chromium often does not expose `AXSelectedText`, so `selection.rs` fails closed. Browsers are documented "best-effort" apps, but browsers are a primary use case.
**Proposal:** fallback capture strategy behind the existing fail-closed guards — e.g. snapshot clipboard → simulated Cmd+C → read selection from clipboard → restore clipboard (reusing `transform_apply`'s clipboard-restore + epoch machinery). Secure-field and permission guards unchanged. Needs a privacy note in the feature doc.

## 2. Transform key silently ignored while dictating
`start_transform_capture: ignored — dictation active` (5× in one session). Correct policy, zero feedback — feels broken.
**Proposal:** reuse the overlay flash path (like hotkey-miss/secure-field flashes) to signal "busy — dictating".

## 3. Rapid transform-key presses surface the main window
Mashing the hold key with a failed/ready popover up can pop the main window open. `set_transform_popover_focusable(true)` → `set_focus()` activates the app; the snapshot/re-hide guard in `set_focusable_internal` loses the race under rapid repeated presses (multiple in-flight focus calls interleave their visibility snapshots).
**Proposal:** make the guard idempotent under interleaving (e.g. remember "main was hidden at flow start" on the session rather than per-call snapshot), or debounce transform-key presses while a popover is already showing.

## 4. (Fixed) failed review blocked dictation
#327 / PR #328 — verified working in the wild: `start_native_recording: auto-dismissed failed transform review`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

georgenijo:
Status at shelf time (2026-07-21, all verified on a built .app, macOS 26, M4):

**Landed & verified (PR #330, merged):**
- Busy flash on refused transform keypress — working
- Dictation key auto-dismisses failed AND ready reviews — working (log: `auto-dismissed transform review`)
- Sticky main-window activation guard — no window pop-ups observed after fix
[...]

PR #346, opened 2026-07-22
fix(transform): visibility snapshot — no resurrecting a closed main window, no force-hiding an opened one (#337)

### LINKED ISSUE #337: Transform: sticky main-window visibility snapshot resurrects a closed window and force-hides an opened one
## Summary

The per-pass "sticky" snapshot of main-window visibility (`transform_main_was_visible`, added in #329) mis-drives the post-`set_focus` re-hide decision in two deterministic ways: it can resurrect a window the user closed mid-pass, and — because supersede paths never clear the snapshot — it can force-hide a window the user deliberately opened.

**Severity:** Low/Medium (unexpected window show/hide).
**Found in:** review of #325/#327/#329, branch `issue/329-mic-prearm` @ `db5d99e`.

## Defect A — closed main window resurrects

The snapshot is write-once per pass (`transform_popover.rs:279-286`, `if was_visible.is_none()`), cleared only in `hide_popover_internal` (`:322-324`). The re-hide decision reads the sticky value (`:419-428`).

### Repro
1. Main Murmur window visible.
2. Select text elsewhere, start a transform → popover shows (snapshot = `Some(true)`).
3. While thinking, close the main window (red button / Cmd-W → hide).
4. At ready, `set_focusable(true)` computes `main_was_hidden = false` → re-hide guard disabled → `set_focus` (`:430`) activates the app and resurfaces the window the user just closed.

The pre-#329 per-call snapshot read "hidden" at focus time and would have re-hidden it.

## Defect B — supersede force-hides a window you just opened

Supersede paths never call `hide_popover_internal`, so the sticky snapshot survives **across** passes (`:280` skips re-record). `keyboard.rs:962-976` (ReviewPending→Idle + clear, no hide) and the Applied-supersede linger early-return (`transform_flow.rs:705-717`, no hide) both start pass N+1 with the pass-N snapshot intact.

### Repro
1. Main window hidden. Select text, transform, reach **ready** — don't approve (snapshot = `Some(false)`).
2. Open the main window via the tray icon.
3. Press right Option again (new capture / supersede).
4. At pass N+1 ready, `main_was_hidden = true` → guard force-hides the window you just opened.

## Suggested fix

Clear `transform_main_was_visible` on every pass boundary, including the supersede paths (`keyboard.rs` supersede branch and the Applied-supersede linger early-return), or move supersede into `start_transform_capture` so it routes through the normal hide/reset. Re-snapshot at each pass's first show.

---

georgenijo:
## Summary

Fixes #337 — two deterministic mis-behaviors of the sticky per-pass main-window visibility snapshot (`transform_main_was_visible`, #329):

**Defect A — closed main window resurrected.** The post-`set_focus` re-hide decision read only the sticky pass-start snapshot; if the user closed the main window while the transform was thinking, the snapshot still said "visible", the guard was disabled, and `set_focus` resurfaced the window they had just closed. The decision is now the pure `should_rehide_main(sticky, live)`: re-hide when the window was hidden at pass start (**keeps the #329 rapid-press guarantee** — a transiently-surfaced window cannot disarm the guard) **or** is hidden right now (closed mid-pass — never resurrect).

**Defect B — stale snapshot force-hides a window you just opened.** The transform-key supersede paths in `keyboard.rs` (Idle / ReviewPending→Idle pass boundary) never route through `hide_popover_internal`, so pass N+1 inherited pass N's snapshot; with a `Some(false)` left over, opening the main window between passes got it force-hidden at the next ready. The pass boundary now clears the snapshot alongside the session clear — gated identically (never under a mid-flight pass) — so each pass re-records at its first popover show.

## Tests

- `should_rehide_main` truth table pinned (4 tests): #329 transient-surface case, #337 closed-mid-pass case, visible-stays case, no-window case. Mutation check: restoring the pre-fix sticky-only semantics fails `rehides_when_user_closed_the_main_window_mid_pass`.
- Defect B's clear is listener wiring (rdev callback → State) — not unit-coverable headless; see checklist.
- Suite: `cargo test -- --test-threads=1` 557 passed · `tsc` clean · `vitest` 272 passed.

## Tier: needs-live-smoke — DO NOT MERGE until a human runs the checklist

Real AppKit window show/hide/focus behavior can't be verified headless.

### Live smoke checklist (built .app)

- [ ] **Defect A repro, fixed:** main window visible → start a transform in Notes → while *thinking*, close the main window (red button) → at *ready*, the popover focuses and the main window **stays closed**.
- [ ] **Defect B repro, fixed:** main window hidden → transform to *ready*, don't approve → open the main window from the tray → press the transform key again (supersede) → at pass 2's *ready*, the main window **stays open**.
- [ ] **#329 regression check:** main window hidden → rapid repeated transform presses (supersede spam) → the main window never visibly flashes/leaks on screen.
- [ ] Normal pass with the main window visible throughout: no unexpected hide.
- [ ] Approve/Undo/Esc paths still behave; popover focus still works for Approve.

Closes #337

### DIFF
--- a/app/src-tauri/src/commands/transform_popover.rs
+++ b/app/src-tauri/src/commands/transform_popover.rs
@@ -402,6 +402,27 @@ pub fn set_transform_popover_focusable(app: tauri::AppHandle, focusable: bool) -
     set_focusable_internal(&app, focusable)
 }
 
+/// Pure re-hide decision for `set_focusable_internal` (issues #329 + #337).
+/// The main window must be re-hidden after `set_focus` when EITHER:
+///
+/// - the sticky pass-start snapshot says it was hidden (`Some(false)`) — the
+///   #329 guarantee: under rapid repeated presses a per-call check can catch
+///   the window transiently surfaced by a previous `set_focus` and would
+///   wrongly disable the guard; or
+/// - it is hidden RIGHT NOW (`currently_visible == Some(false)`) — the #337
+///   defect-A fix: the user closed the main window mid-pass, so focusing the
+///   popover must not resurrect it, even though the pass-start snapshot says
+///   "visible".
+///
+/// `currently_visible` is `None` when there is no main window (nothing to
+/// re-hide) and `true` on an `is_visible()` error (err on the side of not
+/// force-hiding a window the user may be using).
+fn should_rehide_main(sticky_was_visible: Option<bool>, currently_visible: Option<bool>) -> bool {
+    let sticky_hidden = sticky_was_visible.map(|was| !was).unwrap_or(false);
+    let currently_hidden = currently_visible.map(|is| !is).unwrap_or(false);
+    sticky_hidden || currently_hidden
+}
+
 /// Non-command core of `set_transform_popover_focusable` (issue #312 PR-C2).
 pub(crate) fn set_focusable_internal(app: &tauri::AppHandle, focusable: bool) -> Result<(), String> {
     match app.get_webview_window("transform-review") {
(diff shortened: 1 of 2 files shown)


georgenijo:
Fable adversarial review: **zero confirmed defects** — full `should_rehide_main` truth table verified (9/9, #329 sticky-wins preserved by OR-composition), keyboard-boundary clear gating/ordering audited (no clear-vs-record race, no mid-flight snapshot loss, no supersede route bypasses the boundary), mutation kill reproduced. One residual flagged **for the smoke run**: a window surfaced asynchronously by pass N's set_focus after its post-focus re-check is indistinguishable from a user-opened window at pass N+1's record point — checklist item 3 (rapid supersede spam) is the live test for exactly that.

PR #378, opened 2026-07-28
feat: add local accessible theme engine

### LINKED ISSUE #377: Feature: local accessible Theme Engine
## Goal

Deliver the approved local Appearance Theme Engine described in `docs/draft/theme-engine-converged-plan.md` as one tracking issue and one ready-to-merge PR.

## Scope

- [ ] Atomic foundation: versioned `murmur-appearance` storage, strict sanitizer/cache, Sonic resolver, selector-driven Tailwind dark mode, CSP-safe first-paint bootstrap, main/log-viewer synchronization, and main-owned native app theme
- [ ] Appearance Settings page with System/Light/Dark, Sonic reset, and transparent always-dark overlay/transform-review invariants
- [ ] Dependency-free OKLab/OKLCH accessible accent derivation and UI
- [ ] Semantic token-debt migration with complete contrast contracts
- [ ] Background, foreground, and contrast controls
- [ ] Local file import/export via bounded UTF-8 Rust reads and atomic writes; no clipboard use
- [ ] Reference docs, decisions, changelog, and comprehensive automated tests
- [ ] Browser and full native macOS QA across light/dark/system combinations and themed/native windows

## Locked architecture

- Appearance remains separate from `dictation-settings`.
- Main and log-viewer are themed; overlay and transform-review remain transparent, unsynchronized, always-dark glass.
- Main is the only writer and user-change event emitter; user changes are revisioned.
- System appearance changes apply locally per themed window and emit no Tauri event.
- First paint uses a strictly validated write-time resolved token cache loaded by a parser-blocking same-origin script.
- Application-level native `setTheme` is required and owned by main.
- Accent math is dependency-free OKLab/OKLCH with semantic contrast enforcement.
- Dialogs choose paths; bounded Rust commands do UTF-8 reads and atomic writes. Imported caches are discarded and regenerated.
- No cloud, telemetry expansion, marketplace, or clipboard-based theme exchange.

## Canonical plan

`docs/draft/theme-engine-converged-plan.md` is the approved source of truth and implementation record. The four plan tickets are internal delivery phases of this issue.

## Definition of done

A non-draft, mergeable PR with required checks green, full native macOS QA completed, evidence attached, no known blockers, and no unresolved actionable review feedback. The delivery task explicitly forbids merging, auto-merge, tagging, and release actions.

georgenijo:
## Summary

- adds a local, revisioned appearance document with System, Light, and Dark modes; exact Sonic reset; accessible custom accent/background/foreground/contrast derivation; and restart-safe first paint
- keeps the main and Log Viewer windows synchronized, including live macOS appearance changes and native window theme, while preserving the overlay and transform-review windows as always-dark transparent glass
- adds preview-before-apply theme import and authoritative-only export through main-window-gated, bounded Rust file transport
- migrates themed production UI to semantic tokens and adds AST/contrast debt gates, including the newly merged command palette and history workspace
- documents the architecture, commands, events, settings, privacy boundary, canonical plan, and native QA evidence

## Accessibility and privacy

- every mutable/custom theme is resolved against the actual production DOM contrast matrix; 2,000 seeded ordinary themes complete with no emergency fallback, color collapse, or inaccessible pair
- untouched Sonic remains byte-exact, with its narrow static exceptions kept out of supported interactive usage
- import rejects malformed, unsupported, oversized, non-UTF-8, symlink, and non-regular inputs; export is atomic, sibling-temp, parent-fsynced, and `0600`
- theme files never contain resolved cache/revision state and theme exchange never reads or writes the clipboard
- main is the only writer/emitter; Log Viewer is read-only and revision-aware, including explicit rollover

## Validation

- `npm test` — 62 files, 567 tests passed
- `npx tsc --noEmit`
- `npm run build`
- `cargo test -- --test-threads=1` — 640 library tests passed, 5 ignored; all runnable integration suites passed
- deterministic evaluation harness — 16/16 fixtures passed
- workflow policy validation — 42 tests passed
- final macOS release `.app` and `.dmg` bundle build passed (local Xcode 26.5 clang-runtime link workaround; updater artifacts disabled only because the local release signing key is intentionally unavailable)
- independent architecture, accessibility, and privacy/test closure reviews: clean

Native Computer Use QA covered forced/System modes, live OS Light/Dark changes in both themed windows, title-bar agreement, custom extremes, restart persistence, exact reset, valid/malformed/unsupported/oversized import, `0600` export, clipboard sentinel preservation, reduced motion, keyboard/error behavior, the transparent recording overlay, and a real local microphone transcription delivered to history and clipboard. Evidence and the detailed matrix are in `docs/qa/theme-engine/`.

The isolated QA bundle was not granted macOS Accessibility permission, so the global selected-text hold gesture itself was not repeated; the transform-review always-dark boundary is covered by its existing UI tests and the new semantic-token regression gate.

Closes #377


app/src/components/settings/SettingsSection.tsx
```diff
@@ -64,7 +64,7 @@ export function SettingsSection({ title, subtitle, defaultExpanded = true, child
           className={`w-4 h-4 shrink-0 transition-transform duration-200 ${expanded ? 'rotate-180 text-primary' : 'text-on-surface-variant'}`}
           fill="none"
           stroke="currentColor"
-          viewBox="0 0 24 24"
+          viewBox="0 24"
```
georgenijo: **Bug**: `viewBox="0 24"` is invalid — a viewBox needs four numbers (`0 0 24 24`), and per the SVG spec an invalid `viewBox` disables rendering of the element, so the expand/collapse chevron will disappear from every settings section header. Looks like an accidental keystroke during the class migration; it's the only corrupted attribute I found in the whole sweep.

```suggestion
          viewBox="0 0 24 24"
```

app/src/lib/events.ts
```diff
@@ -13,17 +13,17 @@ export const STREAMS: StreamName[] = ['pipeline', 'audio', 'keyboard', 'transfor
...
 
 export const STREAM_COLORS: Record<StreamName, { bg: string; text: string; dot: string }> = {
-  pipeline: { bg: 'bg-stone-200 dark:bg-stone-700', text: 'text-stone-700 dark:text-stone-300', dot: 'bg-stone-500' },
-  audio:    { bg: 'bg-blue-100 dark:bg-blue-900/40', text: 'text-blue-700 dark:text-blue-300', dot: 'bg-blue-500' },
-  keyboard: { bg: 'bg-purple-100 dark:bg-purple-900/40', text: 'text-purple-700 dark:text-purple-300', dot: 'bg-purple-500' },
-  transform:{ bg: 'bg-amber-100 dark:bg-amber-900/40', text: 'text-amber-700 dark:text-amber-300', dot: 'bg-amber-500' },
-  system:   { bg: 'bg-emerald-100 dark:bg-emerald-900/40', text: 'text-emerald-700 dark:text-emerald-300', dot: 'bg-emerald-500' },
+  pipeline: { bg: 'bg-surface-container-high ', text: 'text-on-surface ', dot: 'bg-surface-container-highest' },
+  audio:    { bg: 'bg-primary/10 ', text: 'text-on-surface ', dot: 'bg-primary' },
+  keyboard: { bg: 'bg-primary/10 ', text: 'text-on-surface ', dot: 'bg-primary' },
```
georgenijo: **Question (UX regression?)**: audio/keyboard/transform streams are now visually identical (`bg-primary/10` + `bg-primary` dot), and pipeline's dot is `bg-surface-container-highest`, which is near-invisible against container backgrounds. The log viewer loses the at-a-glance stream discrimination that blue/purple/amber used to provide. If the semantic token vocabulary can't support distinct hues, that's a defensible trade — but it fell out of the migration silently and deserves an explicit DECISIONS.md entry (or a follow-up to derive distinguishable chip colors from the palette).

Nit: the trailing spaces inside these class strings (`'bg-primary/10 '`) end up in the DOM.

app/src/lib/events.ts
```diff
@@ -13,17 +13,17 @@ export const STREAMS: StreamName[] = ['pipeline', 'audio', 'keyboard', 'transfor
...
 export const LEVEL_COLORS: Record<LevelName, string> = {
-  trace: 'text-stone-400 dark:text-stone-500',
-  debug: 'text-stone-500 dark:text-stone-400',
-  info:  'text-stone-700 dark:text-stone-300',
-  warn:  'text-amber-600 dark:text-amber-400',
-  error: 'text-red-600 dark:text-red-400',
+  trace: 'text-on-surface-variant ',
+  debug: 'text-on-surface-variant ',
+  info:  'text-on-surface ',
+  warn:  'text-primary ',
```
georgenijo: **Question**: warn-level log rows now render in the *accent* color rather than the `warning` token — visually indistinguishable from other primary-colored text. `text-warning` exists and is contrast-guaranteed raw on all surfaces per `SEMANTIC_CONTRAST_MATRIX`, so it seems like the intended mapping here. Was `text-primary` deliberate?

app/src/components/RecordingControls.tsx
```diff
@@ -22,7 +22,7 @@ export function RecordingControls({ status, initialized, onStart, onStop, trigge
         <button
           onClick={onStop}
           disabled={!initialized}
-          className="flex items-center gap-2 rounded-xl bg-error px-7 py-3 text-on-primary shadow-lg shadow-error/20 transition-[filter,transform] hover:brightness-95 active:scale-[0.98] disabled:cursor-not-allowed disabled:opacity-50"
+          className="flex items-center gap-2 rounded-xl border border-error/30 bg-error/10 px-7 py-3 text-error shadow-lg shadow-error/20 transition-transform active:scale-[0.98] disabled:cursor-not-allowed disabled:opacity-50"
```
georgenijo: **Design question**: the solid red Stop button (`bg-error text-on-primary`) became a tinted ghost button (`bg-error/10 text-error`). I can see this was forced by the new "no `bg-error` + `text-on-primary`" gate, but Stop is the app's most important CTA and it just got visually much quieter. Was adding a real `on-error` token (so solid error surfaces stay supported) considered instead? If the demotion is intentional, worth a DECISIONS.md note.

app/src/components/PermissionsBanner.tsx
```diff
@@ -152,16 +152,16 @@ export function PermissionsBanner() {
...
                 </button>
-                <p className="text-xs text-amber-600/70 dark:text-amber-400/70">
+                <p className="text-xs text-on-surface">
                   Clears Murmur's stale Microphone entry, then opens System Settings.
                   macOS will re-prompt the next time you record.
                 </p>
                 {micResetError && (
-                  <p className="text-xs text-red-600 dark:text-red-400">
+                  <p className="text-xs text-on-surface">
                     {micResetError}
```
georgenijo, line 168: `micResetError` (and `resetError` at L206) went from red to plain `text-on-surface` — genuine failure text is now visually indistinguishable from the surrounding body copy. I understand the constraint: only `on-surface` is guaranteed on primary tints. But an alternative that keeps semantics is putting this banner on `bg-warning/10` with `text-warning` (a supported pair per your own matrix — `App.tsx`'s dev banner and the AppearanceSettings adjustment notice already use it). The current mix — warning-tinted in some components, primary-tinted here — is the least consistent part of an otherwise disciplined migration.

Related nit: the dismiss button below is now `text-on-surface hover:text-on-surface` — the hover state is a no-op.

app/src/components/settings/AppearanceSettings.tsx
```diff
@@ -0,0 +1,585 @@
...
+          {MODE_OPTIONS.map((option, index) => {
+            const selected = appearance.document.mode === option.value;
+            return (
+              <button
+                key={option.value}
+                type="button"
+                role="radio"
+                aria-checked={selected}
+                tabIndex={selected ? 0 : -1}
+                onClick={() => void appearance.setMode(option.value)}
```
georgenijo: **Unhandled rejection**: `commit()` in the hook deliberately rethrows after setting error state, so `void appearance.setMode(...)` here (and at L306, plus `void appearance.reset()` at L341) produces an unhandled promise rejection whenever a write fails. The hook already surfaces the error via `appearance.error`, so these fire-and-forget call sites should swallow it:

```ts
onClick={() => appearance.setMode(option.value).catch(() => {})}
```

(or stop rethrowing in `commit` for callers that don't need the rejection — the import flow is the only one that does).

app/src/lib/hooks/useAppearance.ts
```diff
@@ -0,0 +1,360 @@
...
+ * owns the application-level native theme.
+ */
+function useMainAppearanceController(): AppearanceController {
+  const initialLoadRef = useRef<ReturnType<typeof loadAppearanceDocument> | null>(null);
+  if (initialLoadRef.current === null) initialLoadRef.current = loadAppearanceDocument();
+  const [document, setDocument] = useState(initialLoadRef.current.document);
+  const documentRef = useRef(document);
+  documentRef.current = document;
+  const initialAppearance = concreteAppearance(document.mode);
+  const initialResolved = resolveTheme(document.theme, initialAppearance);
```
georgenijo: Perf nit: this `resolveTheme` call runs on **every render** of the provider, but its result only feeds `useState` initial values, which are ignored after mount. Post-repair themes hit the fast validation path so it's sub-millisecond in practice, but a lazy/ref-guarded initializer (like `initialLoadRef` just above) would make it free and make the intent explicit.

app/src/lib/appearance/storage.ts
```diff
@@ -0,0 +1,211 @@
...
+  }
+  const document = createAppearanceDocument(
+    sanitized.mode,
+    sanitized.theme,
+    sanitized.revision,
+  );
+  const canonical = JSON.stringify(document);
+  return {
+    document,
+    needsRepair: !sameCache(sanitized.cache, document.cache) || canonical !== JSON.stringify(parsed),
```
georgenijo, line 133: Minor: `canonical !== JSON.stringify(parsed)` is key-order-sensitive, so a semantically identical document with different key ordering triggers a one-time repair write + `appearance-changed` emit. It's benign (self-heals to canonical form) and arguably intended — just worth a comment saying canonical-form drift is deliberately treated as repair, so nobody "fixes" it into a deep-equal later.

app/src/components/FileTranscriptionPanel.tsx
```diff
@@ -68,66 +68,66 @@ export function FileTranscriptionPanel({ addEntry }: FileTranscriptionPanelProps
       <div
         className={`shrink-0 rounded-xl border-2 border-dashed p-8 flex flex-col items-center justify-center text-center gap-3 transition-colors ${
           isDragging
-            ? 'border-stone-500 bg-stone-100 dark:border-stone-400 dark:bg-stone-800'
-            : 'border-stone-300 dark:border-stone-700'
+            ? 'border-outline-variant/40 bg-surface-container  '
+            : 'border-outline-variant/40 '
```
georgenijo: Nit: both branches now share `border-outline-variant/40`, so the drag-over state lost its border emphasis (previously stone-500 vs stone-300) and only the background changes. Also a stray double space in the first branch. Consider something like `border-primary/40 bg-surface-container` for the dragging state to keep the affordance.

app/src/components/FileTranscriptionPanel.tsx
```diff
@@ -68,66 +68,66 @@ export function FileTranscriptionPanel({ addEntry }: FileTranscriptionPanelProps
...
           <path strokeLinecap="round" strokeLinejoin="round" strokeWidth={1.5} d="M3 16.5v2.25A2.25 2.25 0 005.25 21h13.5A2.25 2.25 0 0021 18.75V16.5m-13.5-9L12 3m0 0l4.5 4.5M12 3v13.5" />
         </svg>
-        <div className="text-sm text-stone-600 dark:text-stone-400">
+        <div className="text-sm text-on-surface-variant">
           Drag audio files here, or
         </div>
         <button
           onClick={handlePick}
-          className="px-4 py-2 text-sm font-medium rounded-lg bg-stone-800 text-white hover:bg-stone-700 disabled:opacity-50 disabled:cursor-not-allowed dark:bg-stone-200 dark:text-stone-900 dark:hover:bg-stone-100 transition-colors"
+          className="px-4 py-2 text-sm font-medium rounded-lg bg-surface-container-highest text-on-surface hover:bg-surface-container-highest disabled:opacity-50 disabled:cursor-not-allowed transition-colors"
```
georgenijo: Nit: `hover:bg-surface-container-highest` on an element that is already `bg-surface-container-highest` — the hover state is a no-op (previously stone-800 → stone-700). `hover:bg-surface-container-high` (or an opacity shift) would restore feedback.

app/src/lib/appearance/resolve.ts
```diff
@@ -0,0 +1,639 @@
...
+    tokens['on-surface-variant'] = mixOklab(foreground, tokens.background, 0.22);
+  }
+
+  Object.assign(tokens, appearance === 'light' ? safeTheme.light : safeTheme.dark);
+
+  // Untouched Sonic preserves the exact shipped first-paint/reset fixtures,
+  // but still validates every pair supported by the post-migration UI. Only
+  // the explicit SONIC_LEGACY_CONTRAST_EXCEPTIONS are excluded. Every mutable
+  // path uses the complete matrix below.
+  const untouchedSonic = safeTheme.presetId === 'sonic'
```
georgenijo, line 537: Behavior worth confirming + documenting in the Settings copy: clearing every custom field still leaves `presetId: 'custom'`, which routes through the full mutable matrix — so e.g. `error` shifts away from Sonic's shipped value (the legacy exceptions only apply on this untouched-Sonic path). In other words, "custom with everything cleared" ≠ Sonic; only Reset returns the byte-exact fixture. That's consistent with the plan, but a user who clears all overrides and sees status colors still differ from stock may file it as a bug. A sentence in `docs/features/appearance.md` (and/or near the Reset button) would head that off.

georgenijo (commented):
Reviewed the full diff, with the appearance library, boot script, Rust transport, and hook read line-by-line and every component migration scanned for unintended changes.

Overall this is very strong. The architecture matches the locked plan point for point, the fail-closed instincts are consistent (sanitize-then-resolve, cache never authoritative on load or import, emergency-fallback postcondition in the resolver), and the test depth is exceptional — the boot test executes the committed script artifact itself, the 2,000-seed property test asserts no fallback/color collapse, and the Rust tests pin the atomicity and permission contracts with injected IO. Things I verified specifically:

- **Rust transport**: main-window label gate on both commands, `O_NOFOLLOW` + `symlink_metadata` on read *and* write, 64 KiB bound enforced via `take(MAX+1)` rather than just pre-open metadata, sibling temp + `create_new` + rename with cleanup on every failure path, parent fsync with an explicit failure contract, 0600 for new files with existing-mode preservation.
- **The `dark:` variant re-keying is safe for the always-dark windows** — overlay/transform-review use no `dark:` utilities and no semantic tokens, and neither loads the boot script, so re-keying `dark:` to `data-appearance` cannot change their rendering.
- **No CSS injection surface**: everything reaching `style.setProperty` passes `isHexColor` or comes from the resolver's own hex serializer.
- **Revision protocol**: rollover-to-1 with reserved sentinel, reader stale-revision rejection, snapshot/subscribe race closure, Strict Mode double-mount — all covered by targeted tests.
- Capability grant (`core:app:allow-set-app-theme`) is main-only; command-count doc updates match the two new commands.

Findings are inline: **one real bug** (corrupted `viewBox` in `SettingsSection.tsx` — the settings chevron stops rendering), a set of **migration color-semantics questions** (log-viewer warn level + stream chip collapse, Stop button demotion, error text losing error color in PermissionsBanner) that are user-visible regressions a changelog reader wouldn't expect — if intentional they deserve DECISIONS.md entries — and a few robustness/perf nits.

Verdict: approve once the `viewBox` bug is fixed and the color-semantics questions have explicit answers.

app/src/components/settings/SettingsSection.tsx
georgenijo: Fixed in 0025f90: restored viewBox="0 0 24 24" so the disclosure chevron renders correctly.

app/src/lib/events.ts
georgenijo: Addressed in 0025f90: all five streams now have distinct contrast-safe semantic surface/foreground/marker signatures, event rows render the markers, trailing class whitespace is gone, and the no-new-schema-token choice is recorded in DECISIONS.md. Added constant and rendered-component regressions.

app/src/lib/events.ts
georgenijo: Fixed in 0025f90: warn levels now use text-warning, with both mapping and rendered-row coverage.

app/src/components/RecordingControls.tsx
georgenijo: Addressed in 0025f90 without inventing an unsupported on-error pair: Stop now uses an opaque supported surface, 2px error border, error label, ring/shadow, and safe error-tint hover. The choice is documented and covered by component and native QA.

app/src/components/PermissionsBanner.tsx
georgenijo: Fixed in 0025f90: both permission reset failures are role=alert blocks with error border/text on an opaque surface-container-lowest, and Dismiss now has a real surface hover state.

app/src/components/settings/AppearanceSettings.tsx
georgenijo: Fixed in 0025f90: click, arrow-key, and Reset fire-and-forget actions attach rejection handlers while the hook remains the visible error owner. Added a rejected-controller regression that also verifies the hook error UI.

app/src/lib/hooks/useAppearance.ts
georgenijo: Fixed in 0025f90: the initial concrete appearance and resolver result are now guarded by a mount-only ref snapshot.

app/src/lib/appearance/storage.ts
georgenijo: Documented and tested in 0025f90: byte comparison intentionally treats canonical-form drift as a one-time repair; the new test proves reordered input repairs once and then loads clean.

app/src/components/FileTranscriptionPanel.tsx
georgenijo: Fixed in 0025f90: drag-over now uses a full border-primary treatment with the supported surface background, and the stray whitespace is removed.

app/src/components/FileTranscriptionPanel.tsx
georgenijo: Fixed in 0025f90: Choose Files now moves from surface-container-highest to surface-container-high on hover, restoring visible feedback.

app/src/lib/appearance/resolve.ts
georgenijo: Documented in 0025f90 in both Appearance Settings and the feature guide: clearing individual colors intentionally remains Custom; only Reset to Sonic restores the exact built-in fixtures.

georgenijo (commented):
Re-reviewed the amended head (`0025f90`) against every comment from my previous review. All 11 findings are addressed, each with a regression test or documentation where I asked for one — verified in the diff, not just the description:

1. **`viewBox` bug** — restored to `0 0 24 24`.
2. **Stream chip collapse** — five distinct, matrix-legal treatments (audio keeps `text-on-surface` on the primary tint — the only allowed foreground there; keyboard moved off the tint so it can legally use `text-primary`; transform/system use their status pairs). Distinctness is locked by `events.test.ts` and exercised through real renders in the new `EventSemantics.test.tsx`, and `EventRow` now carries the dot marker too.
3. **`warn` level** — now `text-warning`, with a test pinning it.
4. **Stop button** — opaque lowest surface, `border-2 border-error`, semibold error label, `ring-4 ring-error/10`, error-tint hover: prominent again without the unsupported solid pair, and the no-`on-error`-token rationale is now in DECISIONS.md. Component test locks the treatment.
5. **PermissionsBanner reset errors** — `role="alert"` + `text-error` on an opaque `bg-surface-container-lowest` chip is a clever way to keep genuine error color inside a primary-tinted banner while staying inside the matrix. Dismiss hover fixed.
6. **Unhandled rejections** — `runControllerAction` consumes controller rejections with an intent comment, and the new test covers both `setMode` and `reset` failure paths.
7. **Per-render `resolveTheme`** — now ref-guarded one-time initialization.
8. **Byte-comparison repair** — intent comment added plus a canonical-key-order repair round-trip test.
9. **Drop zone / Choose Files** — drag state regained a real affordance (`border-primary`), hover is no longer a no-op, stray spaces cleaned.
10. **Custom ≠ Sonic after clearing overrides** — documented in `appearance.md` and, better, surfaced directly in the Settings copy next to Reset, with a test asserting the copy.
11. **Error rows** — the old `bg-error/10` row (which quietly put `on-surface` text on an unguaranteed error tint) became `border-l-4 border-error` on an opaque surface — this fixes a latent contrast gap beyond what I flagged.

The QA README also documents a fresh native Computer Use pass on the amended bundle rather than assuming the first round still held — appreciated.

CI on the new head: typecheck, linux, rust-macos, changes, omen all green. No new issues introduced by the remediation. LGTM — this would be an approval, but GitHub blocks approving from the author account.

PR #382, opened 2026-07-28
Fix async single-owner audio initialization lifecycle

### LINKED ISSUE #380: Audio initialization lifecycle: async, single-owner, cancelable recovery
# Audio Initialization Lifecycle — Final Plan

Final convergence of `audio-init-converged-plan.md` and
`audio-initialization-lifecycle-plan.md` (2026-07-28). Supersedes both.

## Incident context

Log session on the MacBook Air, 2026-07-28 14:20–14:29: coreaudiod went slow for
~9 minutes. Device opens took 10–12s against the hard 5s init timeout in
`start_recording` (`audio.rs:183`); six recordings failed silently; timed-out
init threads were forgotten and piled up, contending for the mic (two
`run_audio_capture` lines at 14:22:17.267/.385); an app restart at 14:22:24 did
not help — the wedge was OS-level. The stall moved between phases across
attempts (recording 7: stream build/play; recording 8: enumeration/config),
consistent with a wedged HAL rather than one slow call.

Known live bug found during review: `start_recording` holds the recording-state
mutex across the entire 5s `recv_timeout` (`audio.rs:139–192`), so a stalled
init also blocks `stop_recording` and every other audio call.

## Goal

Make microphone initialization asynchronous, single-owner, cancelable, and
honest about recovery. A timed-out Core Audio thread must remain owned until it
exits — never detached and forgotten. A slow device open becomes a slow start
that succeeds, not a failure to recover from.

## Required invariants

- At most one audio initialization or recording owns the microphone.
- Starting a recording never blocks a Tauri command for five seconds.
- A cancelled or expired attempt can never emit `recording`, audio levels, or
  samples.
- Murmur never returns to `Idle` while an old audio thread is still alive.
- Every spawned audio thread is eventually joined or remains explicitly tracked
  as unrecoverable.
- Recording duration begins when the stream becomes ready, not when
  initialization begins.
- No audio call ever waits on a channel while holding the recording-state
  mutex.

## Generation identity

The existing monotonic `recording_id` is the generation for the whole attempt.
No separate `audio_attempt_id`: with no automatic retry, every attempt is a new
`recording_id`, and a single counter follows the codebase's established
generation pattern. (If a future feature makes one recording span multiple
attempts, introduce the second counter then.)

## State machine

```mermaid
stateDiagram-v2
    [*] --> Idle
    Idle --> Starting: start accepted
    Starting --> Recording: ready and generation valid
    Starting --> Recovering: cancel or hard deadline
    Starting --> Idle: initialization fails and thread exits
    Recording --> Idle: normal stop and teardown
    Recovering --> Idle: abandoned thread exits
    Recovering --> Recovering: late ready is stopped immediately
```

`Recovering` is essential. Returning directly to `Idle` after cancelling a
blocked initialization would allow a second thread and recreate the defect.

## PR 0 — Instrumentation and ownership under the existing synchronous API

Shippable alone, zero UI-visible behavior change, stops the thread pile-up on
day one even if the async work slips.

- **Phase timing**: log enter/exit around device enumeration → config lookup →
  stream build → `play()` → ready signal, tagged with `recording_id`. This
  lands *before* the rewrite so the incident stays measurable and becomes the
  dataset for tuning the deadlines later.
- **Generation-owned init**: on cancellation or timeout, mark the generation
  cancelled; flip `active` off so the callback stops accepting samples and
  stops emitting level events; **retain the thread handle** and reap it when it
  returns; a late `ready` from an expired generation gets an immediate `Stop`
  (closes the orange-dot flicker).
- **Single owner**: while any init is live (pending or draining), a new start
  is rejected fast with a typed error. Never queued — this is push-to-talk with
  immutable per-recording context; a start that fires seconds late records the
  wrong moment with the wrong context.
- **Mutex fix**: release the recording-state lock before waiting on `ready_rx`;
  re-acquire to commit the result. `stop_recording` must be callable while an
  init is stalled.

## PR 1 — Async lifecycle vertical slice

Backend state machine plus the minimum frontend contract, merged together —
the status contract change must be atomic.

### Backend

Replace synchronous `start_recording()` with a supervisor-owned attempt:

- Spawn the audio thread and return immediately; the supervisor owns the thread
  handle from spawn through exit/join.
- Store per-attempt origin, cancellation flag, active flag, command sender,
  timing, and lifecycle state, keyed by `recording_id`.
- Reject starts during `Starting` with `AlreadyStarting`; during `Recovering`
  with `AudioRecovering`.
- The worker reports structured lifecycle messages: phase entered, ready,
  initialization failed, stream stopped, thread exited.
- On `ready`, the supervisor checks the generation:
  - current and uncancelled → transition to `Recording`, set the real start
    time, emit `recording`;
  - cancelled or stale → flip `active` off, send `Stop`, suppress all frontend
    events, remain `Recovering` until the thread exits.
- **Model preparation fires when the start is accepted** (entering `Starting`),
  not on `ready`. Prep is mic-independent; warm-on-record's whole point is
  overlapping model load with other latency, and a slow device open must not
  serialize it. A wasted warm on a cancelled init is acceptable — the idle
  release path already reclaims it.

### Cancellation and deadlines

Separate user experience from cleanup:

- At ~5 seconds in `Starting`: emit a "still connecting" signal. Do not pretend
  initialization failed.
- User activation (toggle, hold release, double-tap) while `Starting`: cancel
  the attempt.
- At ~30 seconds: hard-cancel automatically and report initialization failure
  exactly once. This is a hard initialization deadline, not garbage
  collection — the thread remains owned after it fires.
- After cancellation: remain `Recovering` until the worker exits.
- No automatic retry. The user pressing the key again *is* the retry, and it is
  now safe. No retry loop may hammer a wedged HAL.
- Both deadline values are provisional: revisit them against real phase-timing
  distribution from PR 0 after a week in the field.

### Recording-mode behavior

- Hold release during `Starting` cancels.
- A second toggle/double-tap during `Starting` cancels.
- Silence auto-stop only arms after reaching `Recording`.
- Recording origin is preserved throughout `Starting`.
- Cancelled initialization creates no history entry, statistics, saved audio,
  or transcript.
- Context remains the immutable snapshot captured when the user initiated the
  recording.
- **Device switch or sleep/wake during `Starting` cancels the attempt — never
  adopt a stream across either event.**

### Minimum frontend contract

`useRecordingState` handles `starting`, `recording`, `recovering`, `idle`, and
terminal initialization failure. The overlay shows at least a basic starting
indication in this PR so the backend/frontend contract lands atomically.
Duration and stats derive from the readiness timestamp, never the init start.

## PR 2 — UX and diagnostics polish

Once lifecycle correctness is proven:

- Distinct pulse for `Starting`; after ~5s the indication changes to "still
  connecting."
- Toggle during `Starting` visibly cancels.
- Bounded microphone-failure flash via the existing `useOverlayRuntime` flash
  path.
- While `Recovering`, reject interaction with an actionable message rather than
  appearing idle.
- If recovery exceeds its grace period (thread never exits), surface persistent
  guidance in the existing recording-error surface. Wording must be honest:
  restarting Murmur clears the stuck thread — Rust cannot kill a thread blocked
  inside Core Audio — but macOS audio itself may still need to recover; do not
  promise the microphone will work after restart. (The incident logs show the
  coreaudiod wedge surviving an app restart.)
- Complete the privacy-safe phase telemetry begun in PR 0: `recording_id`,
  phase name and elapsed time, cancellation reason, readiness outcome, stop
  acknowledgment, thread-exit/join confirmation. Release logs continue
  redacting device names.

## PR 3 — Screen-notification coalescing (independent, ship anytime)

- Trailing 100–150ms debounce on screen-parameters-changed.
- Compare the complete display snapshot: notch geometry, monitor bounds,
  origin, scale. Reposition and emit `overlay-geometry-changed` only when the
  snapshot changes — debouncing alone would still process isolated identical
  notifications.
- Log one coalesced event instead of every native notification. (Incident log:
  ~60Hz notification storm for over a second at 13:58, each firing a full notch
  re-detection and reposition.)

## Tests

Injected audio-initialization seam simulating blocking before device lookup,
stream construction, and `play()`. Required cases:

- Start returns promptly while initialization remains blocked.
- A second start never spawns another worker.
- `stop_recording` is not blocked while an init is stalled (mutex fix).
- Cancel during `Starting` enters `Recovering`.
- Late readiness after cancellation is stopped and never emits `recording`.
- `Idle` is not restored until the worker exits.
- Hard deadline emits failure exactly once.
- Stale generations cannot mutate current state.
- Duration begins at readiness.
- Successful retry works after complete teardown.
- Model preparation starts on start-accepted, not on readiness.
- Hold, toggle, double-tap, and silence-auto-stop semantics remain correct.
- Device switch and sleep/wake during `Starting` cancel cleanly.
- No samples or level events escape from abandoned attempts.

Native smoke: rapid overlay clicks and hold-key mashing; logs must never show
overlapping `run_audio_capture` lines. Suspending `coreaudiod`
(`sudo killall -STOP coreaudiod` / `-CONT`) is a documented manual stress test
only — never automated CI — and must demonstrate a 12s device open ending in a
successful recording.

## Acceptance criteria

- No synchronous five-second wait in the start command.
- No forgotten or detached audio thread handles.
- One microphone owner at all times.
- Slow startup is visible and cancelable; a slow-but-successful open produces a
  working recording.
- Late initialization cannot resurrect a cancelled recording.
- Recovery messaging is truthful when Core Audio remains blocked.
- Deadline values re-examined against PR 0 phase-timing data before being
  considered final.
- All recording modes pass the deterministic tests and native
  rapid-interaction smoke testing.

## Resolved disagreements (for the record)

- **`Recovering` state, two-tier 5s/30s signaling, duration-at-readiness,
  mutex fix**: adopted from the lifecycle plan.
- **Model prep at start-accepted** (not on ready): overlap preserved;
  lifecycle plan overruled.
- **Phase timing and measure-then-tune moved to PR 0**: instrumentation
  precedes the rewrite; lifecycle plan overruled.
- **Sleep/wake and device-switch cancellation reinstated** in behavior and
  tests.
- **Restart guidance kept but scoped and reworded** to not promise recovery.
- **Single generation counter** (`recording_id`); no separate attempt ID.
- **PR 0 split kept** (George may veto: folding PR 0 into PR 1 changes nothing
  else in this plan).

georgenijo:
Fixes #380

Summary
- add an app-lifetime audio supervisor with Starting, Recording, Recovering, and owned teardown
- suppress late readiness, samples, and level events for cancelled generations
- add 5s slow-start feedback, a 30s hard deadline, truthful recovery guidance, and readiness-based duration
- cancel in-flight initialization on device changes and sleep or wake
- coalesce screen-parameter notifications against complete display snapshots
- cover dictation and selected-text transform capture with the same non-blocking ownership contract

Validation
- cargo check
- cargo test -- --test-threads=1: 650 passed, 5 ignored; all integrations passed
- npx tsc --noEmit
- npm test: 588 passed
- native macOS smoke against the issue worktree bundle
- reproduced a blocked Core Audio config lookup: command returned promptly, one owner remained tracked, 5s and 30s signals fired, failure emitted once, no recording or overlapping capture escaped, and persistent recovery guidance appeared

The debug app bundle was produced successfully. The packaging command ended after bundle creation because this local environment has the updater public key but no TAURI_SIGNING_PRIVATE_KEY.

## Summary by CodeRabbit

* **New Features**
  * Added clear microphone connection states, including Starting, Still connecting, and Recovering.
  * Microphone startup can be canceled promptly, with visible progress, failure, and recovery guidance.
  * Recording duration now begins when audio is ready.
  * Selected-text transforms show a Connecting phase and offer Retry if audio is released too early.
  * Escape cancellation now works during transform connection.

* **Bug Fixes**
  * Prevented stale audio events and samples from affecting canceled or newer recordings.
  * Reduced unnecessary overlay repositioning during display-change notification bursts.
  * Improved microphone recovery and device-change handling.


app/src-tauri/src/audio.rs
```diff
@@ -68,180 +64,217 @@ macro_rules! build_mono_input_stream {
...
     let devices = host
         .input_devices()
-        .map_err(|e| format!("Failed to enumerate input devices: {}", e))?;
-    let names: Vec<String> = devices.filter_map(|d| d.name().ok()).collect();
-    Ok(names)
+        .map_err(|error| format!("Failed to enumerate input devices: {error}"))?;
+    Ok(devices.filter_map(|device| device.name().ok()).collect())
 }
 
+/// Transform instruction capture retains its ready-before-listening contract,
```
georgenijo: **Transform capture lost its ready-before-listening guarantee.** This comment says the contract is retained, but `start_transform_recording` passes `wait_until_ready: false` — and nothing anywhere passes `true`, so that supervisor path (including its `HARD_INITIALIZATION_DEADLINE + 2s` timeout) is dead code. Previously `start_recording` blocked until cpal readiness, so when `transform_flow` emitted `ReviewState::Listening` the mic was live. Now the popover can show "listening" while the worker is still inside a blocked Core Audio call, and since the `active` gate stays false until accepted readiness, everything the user speaks is silently dropped. Transform also ignores `StillConnecting`/`Recovering` in its `handle_audio_lifecycle`, so unlike dictation there is zero feedback during a slow init — the exact untruthful UI this PR set out to eliminate, moved to the transform surface.

Either pass `wait_until_ready: true` for transform, or gate the Listening transition on the `Ready` lifecycle event and surface a stall cue; and fix this comment either way.

app/src-tauri/src/commands/recording.rs
```diff
@@ -2318,20 +2489,28 @@ pub async fn start_native_recording(
             None,
             None,
         );
-        return Err(e);
+        return match error {
+            AudioStartError::AlreadyStarting => Ok(serde_json::json!({
+                "type": "already_starting",
+                "state": "starting"
+            })),
+            AudioStartError::AudioRecovering => Ok(serde_json::json!({
```
georgenijo, line 2499: **This response can permanently wedge the frontend at `recovering`.** This branch has already emitted `recording-status-changed: "idle"` a few lines up, then returns `state: "recovering"`. The invoke response resolves *after* the events, and `beginRecording` (useRecordingState.ts:209) applies `res.state` to `statusRef` unconditionally — the `staleStartingResponse` guard only covers `'starting'`. So the frontend lands on `recovering` while backend dictation status is `Idle`, and if the recovering attempt is transform-owned, no `recording-status-changed` event will ever clear it. In that state the Start button is disabled, `toggleRecording` ignores input, and overlay clicks return early: dictation is bricked until restart.

Reachable by Escape-cancelling a transform capture and immediately hitting the dictation hotkey while the transform's worker is still unwinding — normally a milliseconds window, but minutes-long in exactly the blocked-teardown scenario this PR targets. Suggest not applying `res.state` for the `audio_recovering`/`already_starting` response types (or reporting a state that matches the actual dictation status).

app/src-tauri/src/commands/recording.rs
```diff
@@ -2358,12 +2538,27 @@ pub async fn stop_native_recording(
...
             DictationStatus::Recording => {
                 dictation.status = DictationStatus::Processing;
-                state.app_state.recording_id.load(Ordering::SeqCst)
+                (DictationStatus::Recording, rid)
             }
         }
     };
+    if status == DictationStatus::Starting {
+        drop(transition);
+        audio_lifecycle::cancel_dictation_initialization(rid, AudioCancelReason::User)?;
```
georgenijo: **Ignored cancel result.** `cancel_dictation_initialization` returns `Ok(bool)`, and `false` means readiness raced in and the attempt is already `Recording` (plausible for a quick hold-release against a fast device: the status read above and the supervisor processing the cancel are not atomic, and `starting_only: true` makes the cancel a no-op once Ready lands). The command still returns `{"type": "recording_cancelled"}` while capture continues live. `cancel_native_recording` handles this correctly via `cancel_dictation_capture` (`starting_only: false`); this branch should either do the same or check the bool and fall through to the normal stop path.

app/src-tauri/src/audio_lifecycle.rs
```diff
@@ -0,0 +1,1456 @@
...
+    let (response_sender, response_receiver) = mpsc::channel();
+    supervisor()
+        .sender
+        .send(SupervisorMessage::Stop {
+            owner,
+            response: response_sender,
+        })
+        .map_err(|_| "Audio lifecycle supervisor is unavailable".to_string())?;
+    response_receiver
+        .recv()
```
georgenijo: **No deadline for the `Stopping` phase.** This `recv()` has no timeout, and `handle_deadlines` covers only `Starting`/`Recovering`. A post-readiness teardown that blocks in Core Audio hangs `stop_native_recording` forever and leaves the supervisor rejecting all future starts with `AlreadyRecording` — with no stalled-guidance signal, unlike the init path. This matches the old `join()` behavior, but the PR's own premise is that macOS teardown can block; a Stopping-phase deadline emitting `RecoveryStalled`-style guidance would close the loop. Fine as a follow-up.

georgenijo (commented):
Reviewed the full diff plus the surrounding arbitration code in `commands/recording.rs` and `transform_flow.rs`. The architecture is strong: the single-writer supervisor actor, generation-owned cancellation, one-shot failure reporting, and the `WorkerFactory`/`LifecycleSink` seams (which make the new tests real state-machine tests) are all clean, and docs/CHANGELOG are thoroughly updated. The display-snapshot coalescing is sound, with a test asserting every dimension participates in the comparison.

Five line comments below: two I'd want addressed before merge (a capture-loss regression on the transform surface, and a frontend state that can wedge dictation until restart), one small targeted fix, and two smaller hardening notes.

Two minors not worth their own threads:

- **Test timing**: the supervisor tests lean on wall-clock sleeps of 10–25ms against millisecond-scale deadlines — `hard_deadline_reports_failure_once_and_retains_owner_until_exit` could flake on a starved CI runner. Polling the sink with a bounded wait would be sturdier than fixed sleeps.
- **Naming**: `audio::start_recording` is now transform-only and takes a `transform_pass_id`; renaming it (e.g. `start_transform_capture_audio`) would stop future callers reaching for it as the generic entry point.

app/src-tauri/src/commands/recording.rs
```diff
@@ -2151,15 +2160,146 @@ pub fn clear_ide_context(
...
+                    serde_json::json!({ "recordingId": recording_id }),
+                );
+            }
+        }
+        AudioLifecycleEvent::Idle => {
+            let mut dictation = state.app_state.dictation.lock_or_recover();
+            if !is_current()
+                || !matches!(
+                    dictation.status,
+                    DictationStatus::Starting | DictationStatus::Recovering
```
georgenijo: **`Idle` doesn't cover a previous status of `Recording`.** If a worker panics mid-recording, `finish_attempt` reports the failure and notifies `Idle`, but this guard leaves dictation status at `Recording` — UI shows recording with a dead mic until the user presses stop, which then self-heals via the empty-samples path. Roughly parity with old behavior, but adding `Recording` here is a one-line hardening. Fine as a follow-up.

app/src-tauri/src/audio.rs
georgenijo: Fixed in 53476d6. Transform now enters `Connecting` after selection capture and transitions to `Listening` only after the exact transform pass owns an accepted audio `Ready` event. The inverse ordering (audio ready before selection) is handled too. `StillConnecting`, `Recovering`, and `RecoveryStalled` now drive truthful popover states/copy, and release-before-ready cancels capture and shows a retryable `audio_not_ready` failure instead of entering Listening late. Added state-machine, integration, and frontend coverage and visually QA’d the new states.

app/src-tauri/src/commands/recording.rs
georgenijo: Fixed in 53476d6. The backend now reports dictation `idle` for `already_starting` / `audio_recovering`, and `useRecordingState` explicitly refuses to apply state from those arbitration responses; lifecycle events remain authoritative. Added the exact regression test where `starting` then `idle` events arrive before the invoke resolves, and verified the frontend stays idle.

app/src-tauri/src/commands/recording.rs
georgenijo: Fixed in 53476d6. The `Starting` stop path now calls `cancel_dictation_capture` for the exact owner with `starting_only: false`, so a concurrent Ready transition cannot turn cancellation into a no-op while the command reports success. The obsolete starting-only helper was removed.

app/src-tauri/src/audio_lifecycle.rs
georgenijo: Implemented in 53476d6 rather than deferring it. `Stopping` now tracks its own deadline and emits `RecoveryStalled` guidance. The stop caller uses a 12-second bounded response wait and returns a truthful blocked-teardown error while the supervisor retains ownership of the worker (no detach / no unsafe restart). A blocking-teardown test proves the guidance event, retained ownership, and eventual join/idle transition.

app/src-tauri/src/commands/recording.rs
georgenijo: Fixed in 53476d6. The `Idle` lifecycle transition now accepts a prior dictation status of `Recording` as well as `Starting` / `Recovering`, so a worker failure or panic after readiness self-heals the backend and UI instead of leaving a dead-mic Recording state.

georgenijo:
Also addressed the two unthreaded review notes in commit 53476d6:

- Replaced fixed 10–25 ms lifecycle-test sleeps with bounded condition polling or FIFO barriers, including the hard-deadline test.
- Renamed the transform-only audio entry point to `start_transform_capture_audio` and updated all callers/docs.

Validation on the pushed commit: Rust unit suite 653 passed / 5 ignored, frontend 591 passed, TypeScript clean, transform integration passed, and the Connecting/Recovering popover states were visually smoke-tested.

georgenijo (commented):
Re-reviewed `53476d67` against all five threads plus the two body minors. Everything is addressed, and CI is green (rust-macos, linux, typecheck).

**1. Transform ready-before-listening — fixed, with a better design than I suggested.** Rather than blocking on `wait_until_ready: true`, the flow gains a real `Connecting` state: `core_start_capture` lands in `Connecting`, and `enter_transform_listening_if_ready` is the only door into `Listening` — driven by the lifecycle `Ready` event *and* healed by a poll of the supervisor's public owner state, which correctly covers the window where `Ready` fires during AX capture before the session exists (the poll in `core_start_capture` catches it; in the retry path the status is `Connecting` before audio starts, so the event can't be missed). Release-before-ready now fails truthfully with `audio_not_ready` and keeps the frozen selection for Retry, and `StillConnecting`/`Recovering`/`RecoveryStalled` all surface in the popover. The atomic `try_transition` makes the two racing callers (supervisor thread vs. command task) emit `Listening` exactly once. The reducer, state-invariant tests (`blocks_recording`, escape targets, cancel-from-every-state), and integration test all grew `Connecting` coverage.

**2. Sticky frontend status — fixed on both sides.** Backend returns `state: "idle"` for the post-allocation `already_starting`/`audio_recovering` responses, and `responseOwnsDictationState` in `beginRecording` stops those response types from overwriting lifecycle events at all — which also safely covers the early-return variants where the response state is legitimate but the events are already authoritative. Regression test added.

**3. Ignored cancel bool — fixed.** `stop_native_recording` now uses `cancel_dictation_capture` (`starting_only: false`) so a readiness race gets discarded instead of left running, and the now-unused `cancel_dictation_initialization` was deleted rather than left around.

**4. Stopping deadline — fixed.** `Stopping` now has `stopping_started_at` + guidance emission in `handle_deadlines`, `stop()` uses a 12s `recv_timeout` with a truthful error, `handle_start` maps `Stopping` to `AudioRecovering` instead of `AlreadyRecording`, and the new `BlockingTeardownFactory` test proves guidance fires without detaching the worker.

**5. Idle-after-Recording — fixed** (guard now includes `DictationStatus::Recording`).

**Minors:** sleeps replaced with a `wait_until` poll helper (plus the FIFO-barrier trick in the stale-Ready test, which is the right way to sequence against an actor); `start_recording` renamed to `start_transform_capture_audio`.

One residual nit, non-blocking: if a post-ready teardown blocks past the 12s stop timeout, `stop_native_recording` errors while backend dictation status stays `Processing`, and `finish_attempt`'s `Stopping` branch doesn't notify the sink when the worker finally exits — so `Processing` lingers until the user cancels. It's a double-failure edge already covered by the stalled guidance; a sink notification from the `Stopping` branch when `stop_response` has no receiver would tidy it up as a follow-up.

Good to merge from my side.

PR #392, opened 2026-07-29
fix: recover from hung Core Audio initialization

### LINKED ISSUE #389: Make microphone initialization recoverable from hung CoreAudio opens
## Problem

On an affected Mac, cpal/CoreAudio can block indefinitely inside `build_input_stream`. The 30-second lifecycle deadline moves Murmur to Recovering but cannot interrupt the owned worker, so dictation remains unavailable until the process exits. Version 0.21.3 detached timed-out workers and allowed unsafe overlap; 0.22.1 correctly retains ownership but has no bounded recovery.

The affected machine has a built-in microphone, iPhone Continuity microphone, and Microsoft Teams Audio virtual device. Release logs prove enumeration and config lookup succeed before `stream_build` blocks. The same underlying blocking behavior existed in 0.21.3.

## Goal

Make the normal macOS dictation path recover to a usable Idle state after an indefinitely blocked microphone open without allowing overlapping in-process CoreAudio workers.

## Scope

- Isolate the potentially unbounded microphone capture lifecycle behind a killable process boundary, or an equivalently safe mechanism that provides a genuine bounded teardown.
- Preserve clipboard-first, local-only audio handling.
- Preserve generation ownership and reject stale readiness/samples.
- Keep normal recording startup overhead negligible and measure it.
- Preserve macOS device selection behavior, including System Default and named devices.
- Add content-free phase and recovery telemetry sufficient to distinguish helper startup, CoreAudio open, deadline termination, and successful restart.
- Provide deterministic tests that simulate an indefinitely blocked open and prove a later attempt succeeds without restarting Murmur.
- Exercise the native app through start, record, stop, transcription, cancellation during startup, and retry.

## Acceptance criteria

- A simulated never-returning audio open is terminated after the deadline and the supervisor returns to Idle.
- A subsequent attempt can record successfully with no stale samples or readiness from the terminated generation.
- Normal successful capture still records and stops correctly.
- User cancellation during initialization settles to Idle within a bounded interval.
- The main app never creates a second in-process CoreAudio worker while an earlier one is blocked.
- Rust tests, TypeScript checks, and native smoke pass.

georgenijo:
Closes #389

## Summary

- release logical microphone ownership immediately after a user cancellation or the 30-second initialization deadline
- keep every abandoned worker callback gate closed and transfer its join handle to one app-lifetime background reaper
- ignore all late events by generation, allowing the next recording attempt without relaunching Murmur
- add a debug-only hung-stream-build fault scenario and update lifecycle documentation

## Important boundary

This does not pretend Rust can interrupt a synchronous Core Audio call. A call that macOS never returns keeps its dormant worker until the OS returns or the app exits. Repeated OS hangs can therefore retain one worker per failed attempt, and a retry can still encounter the same OS problem. The application state is now recoverable and stale callbacks remain closed.

The release success path is unchanged: the fault scenario is compiled only in debug builds, and normal capture still goes directly to the existing cpal worker.

## Proof

- Deterministic lifecycle suite: 10 passed, including immediate retry, stale-generation rejection, callback gating, user cancellation, hard deadline, and event ordering
- Full Rust suite: 654 passed, 0 failed, 5 ignored; all integration suites passed
- Frontend Vitest: 602 passed
- TypeScript: npx tsc --noEmit passed
- Native Tauri debug app: exact device_enumeration -> config_lookup -> blocked stream_build scenario reached Connecting microphone; Cancel returned to Ready with no Recovering state; the next click logged a newly accepted owner without app restart
- Native 30-second deadline flow returned to Ready and accepted the next attempt without restart
- Backend timing sample: cancellation-to-idle 256-351 microseconds; retry acceptance 31-56 microseconds
- Native AX harness cancellation-to-visible-Ready: 894 ms, including UI automation and polling overhead
- vmmap sampled a dormant Rust thread at 2.08 MB virtual stack reservation with 16 KB resident

## Build note

The debug .app bundle was produced successfully. The Tauri command then returned nonzero at its expected updater-signing post-step because no private updater key is present locally.

## Diagnostic note

Repository-wide cargo clippy --lib -- -D warnings still reports 34 pre-existing warnings in unrelated files; this change introduces no Clippy diagnostic in audio_lifecycle.rs.


app/src-tauri/src/audio_lifecycle.rs
```diff
@@ -303,13 +343,71 @@ struct Attempt {
...
+/// single polling thread joins each worker only after macOS lets it return.
+fn abandoned_reaper() -> &'static Sender<AbandonedWorker> {
+    ABANDONED_REAPER.get_or_init(|| {
+        let (sender, receiver) = mpsc::channel::<AbandonedWorker>();
+        std::thread::Builder::new()
+            .name("murmur-audio-reaper".to_string())
+            .spawn(move || {
+                let mut workers = Vec::<AbandonedWorker>::new();
+                loop {
+                    match receiver.recv_timeout(Duration::from_millis(100)) {
```
georgenijo: Once the reaper is initialized (first abandonment ever), this loop wakes every 100 ms for the rest of the app's lifetime, even with zero pending workers. For a battery-conscious menu-bar app that enforces idle timeouts elsewhere, that's a permanent 10 Hz wakeup for what should be a dormant path.

Suggestion: block on `receiver.recv()` when `workers.is_empty()`, and only fall back to `recv_timeout(100ms)` while there are handles to poll:

```rust
let message = if workers.is_empty() {
    receiver.recv().map_err(|_| RecvTimeoutError::Disconnected)
} else {
    receiver.recv_timeout(Duration::from_millis(100))
};
```

app/src-tauri/src/audio_lifecycle.rs
```diff
@@ -1420,13 +1490,45 @@ mod tests {
...
+        assert!(!supervisor.public.is_active());
+
+        let retry_started = Instant::now();
+        let second = AudioOwner::Dictation(31);
+        assert_eq!(start(&supervisor, second).recv().unwrap(), Ok(()));
+        let retry_accepted_us = retry_started.elapsed().as_micros() as u64;
+        assert_eq!(spawn_count.load(Ordering::SeqCst), 2);
+        assert_eq!(cancel(&supervisor, second).recv().unwrap(), Ok(true));
+
+        assert!(recovery_us < 250_000, "recovery took {recovery_us}µs");
```
georgenijo: Hard wall-clock asserts in the default test suite are a CI-flake vector on saturated runners — 250 ms is generous for an in-process channel round-trip, but a loaded machine can blow past it. Since the real value here is the printed benchmark numbers, consider keeping the `println!` and either dropping the asserts, raising the bound well past plausible scheduler noise, or `#[ignore]`-gating the assertion half.

app/src-tauri/src/audio_lifecycle.rs
```diff
@@ -867,51 +953,38 @@ fn handle_deadlines(
...
+            AudioLifecycleEvent::StillConnecting,
+        );
+    }
+    if attempt.as_ref().is_some_and(|current| {
+        current.phase == AttemptPhase::Starting && elapsed >= config.hard_deadline
+    }) {
+        abandon_attempt(attempt, AudioCancelReason::HardDeadline, sink, public, true);
+        return;
+    }
+    let current = attempt.as_mut().expect("attempt was checked above");
```
georgenijo, line 994: Semantics drift worth a follow-up rename: `RecoveryStalled` (emitted just below in the `Stopping` arm) and `recovery_guidance_after` now exclusively mean "stop is stalled" — the recovery phase they were named for no longer exists. The frontend event names (`recording-recovery-stalled`, `audio_recovery_stalled`) carry the same stale vocabulary. Not a bug, but the next reader will hunt for a recovery path that's gone.

app/src-tauri/src/audio_lifecycle.rs
```diff
@@ -209,10 +208,53 @@ impl WorkerFactory for ProductionWorkerFactory {
...
+                            phase,
+                            elapsed_ms: 0,
+                        });
+                    }
+                    let _ = event_sender.send(AudioWorkerEvent::PhaseEntered {
+                        owner,
+                        phase: audio::AudioInitPhase::StreamBuild,
+                    });
+                    loop {
+                        std::thread::park();
```
georgenijo, line 234: Worth a one-line comment that this worker intentionally never exits (simulating a hung Core Audio call), so it's permanently unreapable by design when cancelled in a debug build. Otherwise someone will eventually "fix" the infinite park loop.

app/src-tauri/src/audio_lifecycle.rs
```diff
@@ -449,13 +550,7 @@ fn handle_message(
...
                     public.set_phase(PublicPhase::Stopping);
                 }
-                AttemptPhase::Starting => {
-                    enter_recovering(current, AudioCancelReason::User, sink, public, false);
-                    let _ = response.send(Ok(Vec::new()));
-                }
-                AttemptPhase::Recovering => {
-                    let _ = response.send(Err("Audio is recovering".to_string()));
-                }
+                AttemptPhase::Starting => unreachable!("starting was handled above"),
```
georgenijo, line 561: Minor style: the early-return for `Starting` plus the `expect` plus this `unreachable!` arm is a borrow-checker workaround that could collapse into one clean `match` — e.g. decide the action on `attempt.as_ref()` first, then act on it. Same shape at the tail of `handle_deadlines` (a `match` with one live arm and `_ => {}` could be an `if let`). Fine to leave as-is.

georgenijo (commented):
Reviewed the full supervisor rework against the PR-branch sources, including the event bridge in `commands/recording.rs` and `transform_flow.rs`.

**Verified sound:**
- Event ordering is safe: `LifecycleSink::notify` runs synchronously on the supervisor thread, so the `Recovering → InitializationFailed → Idle` burst completes before the cancel/stop response returns and before any subsequent `Start` is processed. The `Idle` handler in `recording.rs` accepts `Recovering` status, so dictation state lands on `Idle` correctly.
- Late events can't leak: `handle_worker_event` drops events whose owner doesn't match, the abandoned generation's `active` gate is stored false before ownership is released, and `clear_owner` is owner-conditional so a stale abandon can never clobber a newer generation.
- The `expect("attempt was checked above")` calls are genuinely unreachable — every path that takes the attempt returns first.
- The docs honestly state the boundary (one dormant thread per genuinely-hung attempt); tests cover immediate retry, deadline-then-retry with spawn counts, stale-Ready rejection, and callback-gate invariants.

Inline comments below — the reaper idle-poll and the timing asserts are the two I'd address before merge; the rest are follow-up material.

PR #468, opened 2026-08-07
Opt-in "Mirror Captions to NotchPill"

shawngeorgie06:
## What

Adds an opt-in setting that writes the final transcript to a local file so a
separate notch-overlay app can display it. Off by default; when off, nothing
is written and no code runs on the dictation path.

The use case: when you dictate into a window you cannot see — a different
space, a background editor, a terminal behind the browser — there is no way to
check what was actually recognised before it is pasted. Mirroring the caption
lets an always-visible overlay show it word for word.

## How

- `injector::mirror_caption` / `write_caption_to` — atomic (temp file +
  rename) write of `{text, timestamp}` to
  `<data_dir>/local-dictation/latest-caption.json`. Best-effort: every failure
  is swallowed, so it can never affect dictation. Empty/whitespace text is
  skipped.
- The file is written `0600`. It holds a verbatim record of the last thing the
  user said out loud, and the default `0644` makes that readable by every
  account on the machine. Permissions are set on the temp file before the
  rename, so the caption is never briefly world-readable at its real path.
- Setting `mirrorToNotchPill` (default **off**) plumbed through the existing
  delivery path: `settings.ts` / `dictation.ts` → `configure_dictation` →
  `DictationState` → `DeliverySettings` → the inject site.
- Settings UI: a toggle in Delivery, beside the other output options.

## Privacy

Nothing leaves the machine. The write goes to the app's own data directory,
carries only the final text plus a timestamp, is owner-only, and happens only
when the user has explicitly switched it on.

## Tests

- `injector::caption_tests` — valid JSON is written; empty text is skipped.
  Cross-platform, since the existing injector tests are Linux-gated.
- `useSettings.test.tsx` — regression test that the toggle actually reaches the
  backend. The setting was previously missing from the `configure()` trigger
  list, so it saved locally and the backend never heard about it.

Verified locally against v0.27.0: `cargo check` clean, caption tests pass,
`tsc --noEmit` clean, `useSettings` suite 4/4 green, and the feature confirmed
working end to end in a local build.

## Notes

Happy to adjust naming, the file location, or drop the UI string to something
more generic if you would rather it not name a specific third-party app — the
mechanism is not tied to one consumer.

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in setting to mirror the latest dictation transcript in the NotchPill overlay.
  * The setting is saved and applied automatically when changed.
  * Mirrored captions include timestamps, replace previous content, and are removed when mirroring is disabled.
  * Normal text delivery continues unchanged.

* **Documentation**
  * Documented the new caption-mirroring setting and output behavior.

* **Tests**
  * Added coverage for caption handling and settings updates.

### DIFF
--- a/app/src-tauri/src/commands/recording.rs
+++ b/app/src-tauri/src/commands/recording.rs
@@ -957,11 +957,6 @@ async fn run_transcription_pipeline(
     performance_guard.enter(PerformanceStageV1::ClipboardPaste);
     if !text.is_empty() {
         let text_to_inject = text.clone();
-        // Opt-in mirror to NotchPill: best-effort local file write, never blocks
-        // or affects injection. Fires alongside the clipboard write.
-        if delivery.mirror_to_notchpill {
-            injector::mirror_caption(&text_to_inject);
-        }
         let paste_delay_ms = delivery.paste_delay_ms;
         let (tx, rx) = tokio::sync::oneshot::channel::<Result<(), String>>();
         app_handle
@@ -992,6 +987,17 @@ async fn run_transcription_pipeline(
                 let _ = app_handle.emit("auto-paste-failed", paste_hint);
             }
             Ok(Ok(Ok(()))) => {}
+        }
+        // Opt-in mirror to NotchPill: best-effort local file write. It runs
+        // *after* the clipboard write so disk I/O can never delay the primary
+        // output path, and off the pipeline thread so it stays out of paste_ms.
+        // Ownership is re-checked here because the pipeline can be cancelled or
+        // superseded during the paste, and the caption file is last-write-wins:
+        // a stale write would leave NotchPill showing text the user just
+        // abandoned.
+        if delivery.mirror_to_notchpill && !app_state.is_cancelled(recording_id) {
+            let caption = text.clone();
+            tokio::task::spawn_blocking(move || injector::mirror_caption(&caption));
         }
     }
     let paste_ms = t_inject.elapsed().as_millis() as u64;
--- a/app/src-tauri/src/commands/recording.rs
+++ b/app/src-tauri/src/commands/recording.rs
@@ -991,10 +991,15 @@ async fn run_transcription_pipeline(
         // Opt-in mirror to NotchPill: best-effort local file write. It runs
         // *after* the clipboard write so disk I/O can never delay the primary
         // output path, and off the pipeline thread so it stays out of paste_ms.
-        // Ownership is re-checked here because the pipeline can be cancelled or
-        // superseded during the paste, and the caption file is last-write-wins:
-        // a stale write would leave NotchPill showing text the user just
-        // abandoned.
+        //
+        // Ownership is checked before scheduling, which narrows the stale-write
+        // window to the gap between this check and the write landing -- it does
+        // not close it. A cancel arriving inside that gap still publishes this
+        // caption. Deliberate: `app_state` is a borrow and cannot cross into
+        // `spawn_blocking`, and the alternative -- plumbing an owned handle
+        // through purely for a best-effort mirror -- costs more than the
+        // millisecond it would buy. The consequence is bounded: NotchPill shows
+        // a caption the user abandoned until the next one replaces it.
         if delivery.mirror_to_notchpill && !app_state.is_cancelled(recording_id) {
             let caption = text.clone();
             tokio::task::spawn_blocking(move || injector::mirror_caption(&caption));
@@ -1507,7 +1512,16 @@ pub async fn configure_dictation(
         dictation.auto_paste = auto_paste;
     }
     if let Some(mirror) = options.get("mirrorToNotchPill").and_then(|v| v.as_bool()) {
+        let was_enabled = dictation.mirror_to_notchpill;
         dictation.mirror_to_notchpill = mirror;
+        // Turning the mirror off removes what it left behind. The file holds a
+        // verbatim record of the last thing the user said, so leaving it on
+        // disk after they switched the feature off would keep speech around
+        // past consent -- the same reasoning that makes it 0600 in the first
+        // place. Best-effort and errors swallowed, exactly like the write.
+        if was_enabled && !mirror {
+            injector::remove_mirrored_caption();
+        }
     }
 
     if let Some(delay) = options.get("autoPasteDelayMs").and_then(|v| v.as_u64()) {
--- a/app/src-tauri/src/commands/recording.rs
+++ b/app/src-tauri/src/commands/recording.rs
@@ -957,6 +957,11 @@ async fn run_transcription_pipeline(
     performance_guard.enter(PerformanceStageV1::ClipboardPaste);
     if !text.is_empty() {
         let text_to_inject = text.clone();
+        // Opt-in mirror to NotchPill: best-effort local file write, never blocks
+        // or affects injection. Fires alongside the clipboard write.
+        if delivery.mirror_to_notchpill {
+            injector::mirror_caption(&text_to_inject);
+        }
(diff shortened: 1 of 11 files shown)


shawngeorgie06:
Thanks for the review — pushed `266ebf1`. Three of the four findings were real; here's what changed and the one I'm pushing back on.

**Clipboard-first ordering** (`recording.rs`) — correct, and it was the most important one. The mirror ran just before the main-thread dispatch that writes the clipboard, so a slow disk sat in front of the one output path that always has to work. It now runs after the clipboard write returns, on `spawn_blocking`, so it's off the pipeline thread and out of `paste_ms`. That also resolves the separate "move caption I/O off the async pipeline thread" comment.

**Ownership re-check** (`recording.rs`) — correct. The last cancellation checkpoint is before file output, and the pipeline can be cancelled or superseded during the paste. The caption file is last-write-wins, so a stale write would leave NotchPill displaying text the user had already abandoned. Now re-checks `is_cancelled(recording_id)` immediately before mirroring.

**Fail closed on permissions** (`injector.rs`) — correct, and the original code defeated its own purpose. It wrote the temp at the default mode and chmodded to `0600` afterwards, so the verbatim transcript was already on disk at `0644` for the duration of the write. It's now created with `mode(0o600)` in the same `open()` that creates it, and any write or rename failure removes the temp rather than leaving speech content behind. Note that `create_new` makes the mode authoritative but introduces a new failure mode — a leftover temp from a crashed run at the same pid would wedge the feature permanently — so stale temps are cleared first. Two tests added: `write_caption_to_is_owner_only` and `write_caption_to_recovers_from_a_leftover_temp`.

**Rollback field** (`useSettings.ts`) — correct, plain bug. `mirrorToNotchPill` was in the configure trigger list but missing from the revert object, so a failed `configure` left the UI showing a value the backend had rejected, under a message stating that previous settings were restored. Added.

**Declined:** the claim that the `#[cfg(unix)]` path "still does nothing for supported Linux targets." `cfg(unix)` covers Linux — the owner-only policy already applies there. No change.

`cargo test`, `npm test` (636 passing) and `npx tsc --noEmit` are green.

One caveat on my own work, for the record: `write_caption_to_is_owner_only` asserts the final mode, which the old chmod-after-write also satisfied. It locks the contract but would not have caught the original window; that window is closed by construction, not by the test.

georgenijo (changes requested):
Reviewed the current head (post-`266ebf1`). Solid work — this is near merge-ready, and the response to the first review round was right on all four fixes. The declined finding was declined correctly too: `#[cfg(unix)]` does cover Linux, so the owner-only mode already applies there.

## What this gets right

- **Immutable per-recording context** — `mirror_to_notchpill` is snapshotted into `DeliverySettings` via `dictation_context::resolve`, so mid-recording toggles apply to the next session.
- **Clipboard-first** — the mirror runs after the clipboard write returns, on `spawn_blocking`, so disk I/O can't delay the primary output path or pollute `paste_ms`.
- **Generation counter** — `app_state.is_cancelled(recording_id)` matches the existing checkpoints earlier in the same function.
- **Fail-closed write** — `0600` applied at `open()` time rather than chmod-after-write, atomic temp+rename, temp removed on any failure, and a test for the leftover-temp wedge that `create_new` introduces. Appreciated the candid note that the mode test locks the contract but wouldn't have caught the original window.
- **Path contract** — `dirs::data_dir()/local-dictation/` is the app's established convention (telemetry and log shipping already write there), and the `{text, timestamp}` schema matches the planned consumer.
- The `useSettings` configure-trigger + rollback fix closes a real bug class, with a regression test.

## Requested changes

1. **Delete the caption file when the setting is disabled.** Turning the toggle off leaves `latest-caption.json` on disk indefinitely holding the last thing the user said. That doesn't fit the app's fail-closed privacy posture — when `configure_dictation` sees `mirrorToNotchPill` flip to `false`, best-effort remove the file (same swallow-all-errors treatment as the write). Alternatively, document the retention explicitly, but deletion is the better fit.

2. **Docs.** `docs/reference/settings.md` documents every setting, including the "pushed to backend" mapping table — `mirrorToNotchPill` is missing from both. Please add it, plus a line in `docs/FEATURES.md`.

## Notes (non-blocking)

- The "ownership is re-checked here" comment slightly oversells: the `is_cancelled` check happens before `spawn_blocking`, so a cancel landing between the check and the write can still produce a stale caption. The temp filename is also pid-only, so two overlapping writes would share a temp path. Both windows are milliseconds on a best-effort path — fine as is, just worth knowing.
- CI hasn't run on this PR yet (external-fork workflow approval); will approve the run so the suite verifies before merge.

Happy to merge once the two requested changes land.

shawngeorgie06:
Both requested changes are in `0c16b81`, and I fixed both non-blocking notes rather than leaving them.

**1. Delete the caption file when the setting is disabled.** Agreed — speech outliving consent is what the `0600` mode exists to prevent, so leaving the file behind undercut the whole point. `configure_dictation` now removes it on the `true -> false` edge, best-effort with errors swallowed exactly like the write. Orphaned temps go with it: a crashed write leaves one behind and it holds the same speech as the caption. Tests: `remove_caption_clears_the_file_and_any_temp`, and one for a missing directory.

**2. Docs.** `mirrorToNotchPill` added to `docs/reference/settings.md` in all three places — the `Settings` interface block, the settings table, and the pushed-to-backend mapping — and to the Text delivery section of `docs/FEATURES.md`.

**Temp filename collision** — fixed rather than documented. The name now carries a per-write counter alongside the pid, so two overlapping writes from one process can't share a temp path and rename a spliced caption into place. A wrong transcript is worse than a missing one, even on a best-effort path. The leftover-temp test became `write_caption_to_leaves_no_temp_behind`, which asserts no temp of ours survives a successful write.

**The "ownership is re-checked here" comment** — you're right, it oversold. Corrected to say what actually holds: the check happens before scheduling, which *narrows* the stale-write window rather than closing it, and a cancel landing inside that gap still publishes the caption. The comment now also says why it stays that way — `app_state` is a borrow and can't cross into `spawn_blocking`, and plumbing an owned handle through for a best-effort mirror costs more than the millisecond it buys. The consequence is bounded: NotchPill shows a caption the user abandoned until the next one replaces it.

Rust 787 passing, frontend 636 passing, `tsc --noEmit` clean. Ready for the CI run whenever you approve it.

georgenijo (approved):
Re-reviewed `0c16b81`. Both requested changes are in, and both non-blocking notes were fixed rather than papered over. Approving.

**Delete on disable** — verified the edge logic, not just the diff: `buildConfigureOptions` always carries `mirrorToNotchPill`, so every configure call reaches the check, and `was_enabled` compares against live `DictationState`, so the delete fires exactly on the `true -> false` transition and never on unrelated setting changes. The app-restart case resolves correctly too (persisted `true` re-arms state on launch, so a later disable still hits the edge). Sweeping orphaned temps in the same pass was the right call — they hold the same speech.

**Temp-name counter** — correct fix, and it also quietly retires the old pid-reuse wedge: a crashed run's leftover can now collide with at most one write (same pid after restart, seq 0), which fails closed and self-heals on the next write. "A wrong transcript is worse than a missing one" is exactly the right priority for this feature.

**Comment** — now says what the code actually guarantees, including why the residual window stays open and what it costs. That's what a maintenance comment should be.

**Docs** — all three spots in `settings.md` plus FEATURES.md, matching the surrounding entries' level of detail.

One footnote, not blocking: `remove_mirrored_caption` does synchronous fs work inside the async `configure_dictation` handler. It's a handful of syscalls on a settings path, so it doesn't matter in practice — just noting it wasn't overlooked.

CI still hasn't run (first-time-contributor workflow approval); will approve the run and merge once it's green.

Issue #486, opened 2026-08-07
Reliability: recordings never fail — zero stuck states


georgenijo:
## Goal

Dictation must be boringly reliable: pressing the hotkey always either records or shows a clear, recoverable error. No stuck states that require an app restart.

## Finish line (measurable, from fleet telemetry)

- Zero occurrences per week of a state that requires restarting the app (stuck Recovering, lingering Processing).
- 99.5% of key-presses reach first PCM within 400ms (excluding permission prompts).
- Every failure path surfaces a user-visible, actionable status.

## Child issues

- #388 — stuck stream_build leaves app in Recovering until restart
- #384 — status lingers at Processing after timed-out Stopping teardown
- #446 — on-device capture-health signal when primary backend chronically fails over
- #447 — scheduled regression watch over shipped capture-startup metrics
- #472 — capture-helper spawn churn portion

Umbrella/tracking issue — close when the finish line holds for two consecutive weeks of production telemetry.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

georgenijo:
Audit against current `origin/main` at `5c1cdad6c575d5fa8595c09623ff972f13653b99` (2026-08-11), GitHub timelines, and the privacy-safe fleet aggregate:

**What is actually in main**
- #384, #388, #446, and #447 are implemented by cumulative merged PR #506. Their individual PRs #489–#492 were closed unmerged; #506 is the merge that matters.
- #472's dominant reliability work is merged: PR #493 removed the 60-second log-shipper enumeration churn and PR #498 migrated the stale 50 ms paste default. Residual shared inventory is #534 (open); PR #530 only handles capture-owned deferral and is still open/unmerged.
- Measurement/diagnostic follow-ups remain incomplete: #511 → PR #526 open/unmerged; #514 → PR #532 open/unmerged; #535 and #536 are open.

**What production evidence currently proves**
- The fleet capture report generated `2026-08-11T04:44:16Z` is `healthy` with 0 alerts, 5 scanned installs, and 0 malformed lines.
- For versioned v0.30.1 data from `2026-08-10T16:41:47Z` through `2026-08-11T05:11:50Z`, there are 51 dictation readiness successes: 51/51 at or below 400 ms, p50 214 ms, p95 283 ms, max 336 ms; no stable lifecycle-failure/capture-failure/stall event was counted in that cohort.

**Why this cannot close**
- The v0.30.1 window is under 13 hours, not two complete weeks.
[...]

PR #490, opened 2026-08-07
fix: clear Processing after late audio teardown

### LINKED ISSUE #384: Audio: dictation status lingers at Processing after a timed-out Stopping teardown
Follow-up from the #382 re-review (residual nit, non-blocking there).

## Scenario

A post-readiness teardown that blocks in Core Audio for more than the 12s `STOP_RESPONSE_TIMEOUT`:

1. `stop_native_recording` has already committed `DictationStatus::Processing`, then `audio_lifecycle::stop()` times out and the command returns an error. The `map_err` path emits `recording-status-changed: "idle"` but does **not** reset the backend dictation status — it stays `Processing`.
2. The supervisor's attempt stays correctly tracked in `Stopping` (this is by design — the worker is never detached), and the stalled-teardown guidance fires at 10s.
3. When the worker finally exits, `finish_attempt`'s `Stopping` branch sends the samples to a `stop_response` whose receiver was dropped at the timeout, and — unlike the `Starting`/`Recovering`/`Recording` branches — never notifies the sink.

Result: backend dictation status lingers at `Processing` with no event to clear it. The UI shows idle (from the error-path emit), so the next start attempt returns the `"currently processing"` busy response, the frontend applies `state: "processing"`, and the user has to Escape-cancel (or restart) to get back to a working state.

## Suggested fix

In `finish_attempt`'s `AttemptPhase::Stopping` branch, when `stop_response.take()` has no live receiver (send fails), notify the sink with `AudioLifecycleEvent::Idle` so the dictation bridge can run its normal `Processing`-aware reset (the `Idle` handler would need `Processing` added to its status guard for this path). Alternatively, have the `stop()` timeout path itself schedule the status reset once the owner reports exit.

Either way the invariant to keep: the worker stays owned and joined — this is only about the *dictation status* not being reset after the stop caller has already given up.

## Reproduction / test seam

`BlockingTeardownFactory` in `audio_lifecycle.rs` tests already models a blocked teardown; extending that test to drop the stop receiver before opening the gate, then asserting a sink `Idle` notification, covers the supervisor half. The dictation-bridge half is a `handle_audio_lifecycle` unit case.

Requires a >12s blocked post-ready teardown *and* a retry — an edge-of-edge double failure, hence follow-up rather than a #382 blocker.

georgenijo:
## Summary
- publish lifecycle Idle when a stopped worker has exited and joined after its stop caller already timed out
- allow that current-owner Idle notification to clear dictation Processing and the keyboard processing gate
- prove the blocked worker remains exclusively owned until join, then a fresh recording succeeds
- document the late-teardown status contract

## Failure mode fixed
stop_native_recording commits Processing before waiting for the audio supervisor. If the stop response exceeds its timeout, the caller disappears while the supervisor correctly retains the Stopping owner. When the worker later exited, the response send failed silently and no lifecycle event reset backend Processing, leaving the next recording busy until Escape or restart.

The fix does not detach or overlap workers: Idle is published only after ThreadExited has been joined and only when the stop result cannot be delivered to its original caller.

## Production evidence
Current fleet MacBook logs show recent teardown-and-resample operations consistently completing in roughly 8-22 ms and no retained 12-second stop-timeout occurrence. This remains a latent double-failure guard, covered through the deterministic BlockingTeardownFactory seam.

## Verification
- cargo check --all-targets
- cargo test -- --test-threads=1 (751 unit passed, 5 ignored; 12 capture-helper integration passed; 21 sidecar integration passed; transcription and transform-flow integration passed)
- npx tsc --noEmit
- npm test (70 files, 649 tests)
- native debug app opened Ready; two consecutive recording attempts were accepted and returned to Ready with the local actionable microphone error, with no lingering Processing gate
- debug app and DMG were produced; packaging stopped only at the expected missing release signing private key

Closes #384
Related to #486

### DIFF
--- a/app/src-tauri/src/audio_lifecycle.rs
+++ b/app/src-tauri/src/audio_lifecycle.rs
@@ -869,8 +869,16 @@ fn finish_attempt(attempt: &mut Option<Attempt>, sink: &dyn LifecycleSink, publi
     match finished.phase {
         AttemptPhase::Stopping => {
             let samples = take_samples(&mut finished);
-            if let Some(response) = finished.stop_response.take() {
-                let _ = response.send(Ok(samples));
+            let response_delivered = finished
+                .stop_response
+                .take()
+                .is_some_and(|response| response.send(Ok(samples)).is_ok());
+            if !response_delivered {
+                sink.notify(
+                    finished.app_handle.as_ref(),
+                    finished.owner,
+                    AudioLifecycleEvent::Idle,
+                );
             }
         }
         AttemptPhase::Starting => {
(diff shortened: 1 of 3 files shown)


georgenijo:

Issue #487, opened 2026-08-07
Speed: minimize press-to-paste latency


georgenijo:
## Goal

Time from releasing the hotkey to text pasted is the single number that defines how polished Murmur feels. Measure it, put it on the fleet dashboard, and drive it down.

## Finish line

- Press-to-paste (stop → text in target app) measured end-to-end in production telemetry and visible on the fleet dashboard.
- A target set from baseline data (e.g. p50 under 1s, p95 under 2s for a typical utterance on the default model) and met.

## Child issues

- #340 — transform latency (warm-on-arm, skip AX retry on browsers, token streaming)
- #472 — stale paste delay portion

## Candidate work beyond existing issues

- Audit the stop → transcribe → transform-pipeline → clipboard → paste path for serial waits that could overlap.
- Longer term: streaming/partial transcription so text appears while speaking (separate issue if pursued).

Umbrella/tracking issue.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

georgenijo:
Audit against current `origin/main` at `5c1cdad6c575d5fa8595c09623ff972f13653b99` (2026-08-11), merged history, and privacy-safe production metrics:

**Delivered work**
- #340 is in main through cumulative merged PR #506; its individual PR #495 was closed unmerged. Main contains transform warm-on-arm, the bounded browser AX retry reduction, and streamed rewrite preview.
- #472's speed item is fixed in main: PR #498 migrates an unversioned exact legacy 50 ms paste delay to 0 once. The v0.30.1 production paste-stage proxy now has p50 4 ms / p95 48 ms (n=39), versus the earlier v0.28 cohort's 55/83 ms.
- #472's remaining helper work is reliability rather than press-to-paste latency: shared inventory #534 is open, and capture-owner deferral PR #530 is still open/unmerged. #535 and #536 also remain open and affect trustworthy lifecycle/diagnostic coverage.

**Available baseline proxy**
- For non-empty v0.30.1 completions in the versioned fleet window `2026-08-10T16:41:47Z`–`2026-08-11T05:11:50Z`, existing `pipeline.dictation_completed.total_ms` reports p50 183 ms / p95 471 ms (n=39).
- Restricting only by content-free utterance duration (1–15 s) gives p50 172 ms / p95 239 ms (n=30).

**Why this is not the finish line**
[...]

PR #495, opened 2026-08-07
perf: reduce transform time to visible rewrite

### LINKED ISSUE #340: Transform: reduce end-to-end latency (warm-on-arm, skip AX retry on browsers, token streaming)

georgenijo:
## Summary

- prewarm the signed local-LLM helper/model as soon as the transform hold is armed, with pass-scoped cancellation so abandoned loads cannot block dictation
- skip the two 250 ms AX warm-up retries for known Chromium browser bundle IDs while retaining the first fail-closed security check and bounded clipboard fallback
- upgrade the local-LLM protocol to v3 with bounded, contiguous output chunks and render a review-window-only cumulative preview; the final validated result remains the sole approvable/applyable proposal

Production dogfood telemetry on the fleet MacBook showed both recent fallback captures exhausting three AX attempts before another 312–321 ms clipboard fallback, roughly 0.8 s from first attempt to terminal capture. This change removes only those browser retry gaps.

## Verification

- `cd app/src-tauri && cargo check --all-targets`
- `cd app/src-tauri && cargo test -- --test-threads=1` (750 unit tests passed; integration suites including 24 sidecar tests passed)
- `cd app && npm test` (70 files / 652 tests passed)
- local-LLM sidecar crate check and rebuilt bundled helper
- native `tauri:dev` launch plus Playwright visual check of the expanded streaming Thinking state

Closes #340

Umbrella: #487


georgenijo:

PR #492, opened 2026-08-07
feat: schedule capture-startup regression watch

georgenijo:
## Summary
- stamp accepted receiver events with the bounded app-version header already shipped with each batch
- add an hourly, stdlib-only systemd watch over retained capture-startup telemetry
- aggregate per install/version p50/p95, backend/setup-step timeout histograms, fallback/failure counts, and ready recordings per completed attempted session
- alert on >2x p50 regressions with at least five samples per cohort and repeated completed zero-ready sessions
- surface the atomically written report on the protected receiver dashboard

## Failure mode fixed
The v0.24.2 zero-ready install and v0.25.0 startup regression were found only through manual JSONL greps. This makes the same privacy-stripped signals a scheduled operational check. Idle launches and the currently open session cannot create zero-ready alerts; historical unversioned events remain `unknown` and never enter version comparisons.

## Privacy and bounds
- no new client-collected fields; version attribution comes from the existing ingest header
- client-supplied `ingest_app_version` is removed so it cannot spoof cohorts
- backend/setup-step labels are allowlisted and source summaries/content are excluded from reports
- scanning is line-by-line with 500 readiness samples per cohort and 64 explicit versions per install

## Verification
- `python3 -m unittest discover -s tests` (108 passed)
- `cd app/src-tauri && cargo check --all-targets`
- `cd app/src-tauri && cargo test -- --test-threads=1` (749 unit passed, 5 ignored; integrations passed)
- `cd app && npx tsc --noEmit`
- `cd app && npm test` (649 passed)
- real MacBook dogfood telemetry confirmed stable `startup_baseline`, `audio.capture_ready`, `startup_ms`, and capture fallback/timeout event shapes; current starts were ~198-276 ms

Deployment and rollback commands for the receiver script, service, and timer are documented; production is intentionally not changed before review/merge.

Closes #447
Related to #486

## Summary by CodeRabbit

* **New Features**
  * Added hourly monitoring for capture startup performance and regressions.
  * Added dashboard health and alert indicators based on capture-watch reports.
  * Ingested events now record validated app-version information when available.
  * Added metrics for startup timing, timeouts, fallbacks, failures, and session readiness.
* **Documentation**
  * Added deployment, rollback, retention, privacy, and alert-threshold guidance.
* **Tests**
  * Added coverage for regression detection, malformed events, version handling, and dashboard alerts.

PR #491, opened 2026-08-07
feat: surface on-device capture startup health

georgenijo:
## Summary
- derive a bounded five-recording microphone startup signal from existing on-device capture events
- flag chronic backend fallback or a median start-to-audio time of at least two seconds
- surface healthy, degraded, and insufficient-data states in Diagnostics > Performance without changing capture behavior
- accept only stable AUHAL/CPAL labels and content-free event fields

## Failure mode fixed
Production dogfood history showed repeated AUHAL-to-CPAL rescue episodes, but users had no durable on-device indication that the primary backend was chronically failing. The performance-run schema does not currently store capture startup fields, so this reads the bounded cross-launch event history already emitted by the capture path and correlates fallback to readiness by owner and time window.

## Verification
- `cd app/src-tauri && cargo check --all-targets`
- `cd app/src-tauri && cargo test -- --test-threads=1` (749 unit passed, 5 ignored; integrations passed)
- `cd app && npx tsc --noEmit`
- `cd app && npm test` (657 passed)
- focused component tests render healthy, degraded, and insufficient-data copy
- production telemetry checked on the fleet MacBook; recent starts were healthy (~198-276 ms), while retained historical logs contained repeated AUHAL-to-CPAL fallback episodes

Native visual attachment was blocked because the separate health-sweep dev app owns the shared dev single-instance slot; that process was left untouched per coordination instructions.

Closes #446
Related to #486

PR #489, opened 2026-08-07
fix: make blocked audio recovery exercise the killable helper

georgenijo:
## Summary
- move the debug stream-build hang fault into the real capture worker process, so the production terminate/kill/reap boundary owns the simulated blocked HAL call
- keep device enumeration from consuming the one-shot fault sentinel, allowing the first capture start to exercise recovery
- remove the obsolete in-process parked-thread path and document the fault contract

## Failure mode fixed
The existing hang_stream_build fault parked an app-process Rust thread forever. That bypassed the process-isolated capture path shipped by #421 and left the supervisor in exclusive recovery, so dogfood acceptance could not prove that a blocked stream build was killable or that a fresh start was accepted without restarting Murmur.

## Production evidence
Fleet MacBook logs retained from 2026-07-30 through 2026-08-07 contain 200 recent audio.capture_ready events (224 ms average, 591 ms maximum, two above 400 ms). The older blocked/reaper warnings stop after the capture-helper rollout; later backend failures do not leave capture stuck in recovery. This PR turns the rollout acceptance case into a faithful helper-process fault path.

## Verification
- cargo check --all-targets
- cargo test -- --test-threads=1 (750 unit passed, 5 ignored; 12 capture-helper integration passed; 21 sidecar integration passed; transcription and transform-flow integration passed)
- npx tsc --noEmit
- npm test (70 files, 649 tests)
- debug app bundle produced; packaging stopped only at the expected missing release signing key
- native smoke with hang_stream_build_once: first owner entered StreamOpen, cancellation hard-stopped and reaped the helper in 257 ms, UI returned to Ready, and a second recording owner was accepted immediately without process overlap

Closes #388
Related to #486

PR #440, opened 2026-08-03
docs: add Cursor Cloud Linux setup notes to AGENTS.md

georgenijo:
## Summary

Sets up the Cloud Agent development environment for Murmur and documents the non-obvious Linux toolchain requirements. Murmur is a **macOS-only** Tauri app, but the Cloud VM is **Linux (Ubuntu 24.04, x86_64)**, so this establishes what is realistically runnable on Linux — the React frontend and the Rust `cargo test --lib` suite — exactly matching what CI runs on its `ubuntu` jobs.

The only code change is a new `## Cursor Cloud specific instructions` section in `AGENTS.md`. No application code was modified.

## What was set up (persists in the VM snapshot)

- Frontend deps via `npm --prefix app install` (Node 22).
- Rust `stable` (1.97.1) set as the rustup default — the VM was pinned to 1.83, but the `rdev` git fork requires `edition2024` (Rust ≥ 1.85).
- CUDA 12.8 toolkit (`nvcc`, `cudart-dev`, `libcublas-dev`) + a `libcuda.so.1` driver stub on the loader path, because `whisper-rs` is built with the `cuda` feature on Linux and the VM has no GPU/driver.
- `g++-14` / `libstdc++-14-dev` — the default clang selects the gcc-14 toolchain dir, and whisper.cpp's CMake build fails with `cannot find -lstdc++` without it.
- Tauri/audio apt deps: `libwebkit2gtk-4.1-dev libappindicator3-dev librsvg2-dev patchelf libxdo-dev libasound2-dev`.

## Verification

| Surface | Command | Result |
|---|---|---|
| Frontend typecheck | `cd app && npx tsc --noEmit` | pass |
| Frontend tests | `cd app && npm test` | 623 passed (66 files) |
| Frontend build | `cd app && npm run build` | built |
| Frontend run | `cd app && npm run dev` (http://localhost:1420) | UI renders + navigable |
| Rust tests | `cd app/src-tauri && cargo test --lib -- --test-threads=1` | 676 passed, 0 failed, 5 ignored |

### Hello-world demo
Navigated the first-launch onboarding wizard end to end (Welcome → Microphone → Accessibility → Model → "You're all set") into the main dashboard. The `invoke` error overlay in the recording is expected — the macOS-only Rust backend does not run on Linux, so Tauri IPC calls fail while the UI still renders and navigates.

[...]

PR #505, opened 2026-08-07
chore: complete sequenced codebase health sweep

### LINKED ISSUE #439: Harden updater state ownership and release-manifest policy
## Problem

The released updater flow in v0.24.2 has reliability gaps across both runtime state ownership and release-manifest generation.

### 1. Download/install is not single-owner

`useAutoUpdater.startDownload()` has no in-flight guard. It awaits `getUpdateInstallEnvironment()` before transitioning to `downloading`, leaving the Update Now action active long enough for a double-click to enter the path twice. A targeted regression test against v0.24.2 reproduces two environment checks from two immediate calls; both calls can then reach `downloadAndInstall()`.

Manual or lifecycle update checks are also gated only by `isCheckingRef`, not by download/install activity, so they can replace `UpdateStatus` and `updateRef` while an install is in progress.

### 2. Required-update policy cannot be published

The app reads an optional `min_version` from `latest-v2.json`, but `scripts/release_artifacts.py manifests` has no input for it and never writes it. The release workflow has no corresponding policy input, and the live v0.24.2 manifest contains no `min_version`. The documented Required Update flow therefore cannot be activated through the normal release pipeline.

The runtime also collapses “field absent” and “manifest policy fetch failed” to `null`, silently treating both as optional.

### 3. Manifest notes can drift from the final release body

The v0.24.2 updater manifest contains the generated `Other Changes` body, while the GitHub release page contains a later polished `Fixes` summary. Update Available and What's New consequently show different copy from the public release.

## Expected behavior

- Exactly one update operation owns check/download/install state at a time.
- Repeated Update Now actions are idempotent while verification, download, install, or relaunch is in progress.
- Manual, timer, focus, visibility, and wake-triggered checks cannot overwrite an active install or its error state.
- The release pipeline can intentionally publish a validated optional `min_version` policy.
- Policy absence is distinguishable from policy-fetch failure, with an explicit failure rule that does not silently downgrade a configured required update.
- The updater manifest and final GitHub release expose the same release notes at publication time, with a documented policy for later edits.

## Acceptance criteria

- [ ] Add an explicit single-owner operation guard/state machine around check and install paths.
- [ ] Disable or safely no-op all duplicate install entry points while an install owns the operation.
- [ ] Add hook regression tests for double-click, manual-check-during-install, and lifecycle-check-during-install.
- [ ] Add a validated optional `min_version` argument to manifest generation and release promotion.
- [ ] Reject malformed `min_version` and values newer than the release version.
- [ ] Test manifests both with and without `min_version`.
- [ ] Make policy fetch results typed (`present`, `absent`, `unavailable`) and test the chosen failure behavior.
- [ ] Verify manifest notes match the release body before publication and document post-publication edit behavior.
- [ ] Run frontend tests/TypeScript, release-script tests, Rust checks/tests, and a native updater UI smoke test using a non-production mock seam.

## Evidence

- Released code: `app/src/lib/hooks/useAutoUpdater.ts` in v0.24.2
- Manifest generator: `scripts/release_artifacts.py::write_updater_manifests`
- Live release manifest: `releases/latest/download/latest-v2.json`
- Live release: v0.24.2

### LINKED ISSUE #476: Audio: stop spawning a capture helper every 60 s from the log shipper
## Problem

The diagnostic log shipper spawns a full signed capture-helper **process** every 60 seconds, for the entire time the app is running, on an idle machine.

`log_shipper.rs:376`:
```rust
let snap = tokio::task::spawn_blocking(audio_state).await.unwrap_or_default();
if !snap.is_empty() && last_snapshot.as_deref() != Some(snap.as_str()) { ... }
```

`audio_state()` (`log_shipper.rs:284-289`) calls `crate::audio::list_input_devices()`, which spawns a signed helper process, handshakes, sends one `Enumerate`, and kills the child (`audio.rs:400-450`). That call sits in the shipper loop with `TICK_SECS = 60` (`log_shipper.rs:18`). Log shipping is on by default in release builds.

The comment above it — "Event-driven device/state snapshot: POST only when it changes" — is accurate about the **POST** and misleading about the **spawn**: change detection happens after `audio_state()` has already returned, so the spawn is unconditional per tick.

## Evidence

From the v0.28.0 dogfood install (see #472):

- 1,002 distinct capture IDs, only **187** ever received a `start` — 815 helper lifecycles served no recording
- **1,472 of 1,511** orphan spawn events landed more than 10 s from any recording
- Hours with **zero** recordings still show 22–53 spawns
- Works out to roughly **1,440 spawns/day** per running install
- Each spawn costs ~21 ms mean signature verification (p95 26 ms, one 2,108 ms outlier)

## Why this matters

Not latency — these are off the dictation hot path, so fixing this makes dictation no faster. It is a **reliability** concern: cycling CoreAudio open/close ~1,440 times a day is the tap-churning pattern implicated in the coreaudiod stalls this whole subsystem was rebuilt to survive.

## Proposed fix

Make the device snapshot genuinely event-driven — register a `kAudioHardwarePropertyDevices` listener and only recompute on change — or cache the result behind a long TTL so an idle install stops cycling CoreAudio.

## Acceptance

An idle install (app running, no recordings, Settings closed) spawns **no** capture helper processes over a 30-minute window.

### LINKED ISSUE #477: CI: make ci-pass an honest required check
## Problem

CI is advisory. Nothing it reports can block a merge, and the sentinel built to be the required check has a hole in it.

Three separate defects:

**1. `main` has no required status checks.** Verified via API: `required_status_checks: null`. The `ci-pass` job (`ci.yml:200-238`) exists specifically to be that single required check, and nothing references it. A red CI run does not block a merge today.

**2. `ci-pass` omits `visual-regression` from its `needs`.** `ci.yml:201` reads `needs: [changes, typecheck, rust-macos, linux]`, but `visual-regression` is a real job at `ci.yml:92-112` (added in 982e557, "test: lock canonical Murmur UI states"). Requiring `ci-pass` as-is would create a gate that a red visual golden test passes straight through.

**3. The existing review requirement is already a dead letter.** Verified: `enforce_admins: false`, `required_approving_review_count: 1`, `require_last_push_approval: true`. For a solo maintainer that combination is unsatisfiable — you cannot approve your own last push — which means every sub-hour merge is already an admin override. Adding a second rule to a regime that is routinely bypassed accomplishes nothing.

## Proposed fix

In order:

1. Add `visual-regression` to `ci-pass`'s `needs` and to its result loop
2. Set `required_approving_review_count` to `0` — make `ci-pass` the honest gate rather than pretending a review happens
3. Enable `strict` (require branches up to date before merging)
4. Require the `ci-pass` context

## Follow-on: only then is the duplicate push run safe to remove

CI runs on both `pull_request` and `push` to main (`ci.yml:3-6`) — 8 runs / 20 runner-minutes per 100 runs. That looks like pure duplication, but with `strict` off, a PR green on a stale base can merge and break main, and **the push run is currently the only thing that catches it**. Do not remove it before step 3.

## Verification note

`ci-pass`'s `if: always() && ...` logic is otherwise correct — it handles the intentionally-skipped release-bump path without deadlocking. The guard does not need changing.

### LINKED ISSUE #478: Capture worker: unit-test the pure framing and ring-buffer logic
## Problem

The capture worker sidecar carries the densest fix history in the repo — including a crash that shipped in release v0.24.0 (#428) — and its Rust logic has no unit tests.

`app/src-tauri/sidecars/capture/`: `main.rs` (416 lines) and `production.rs` (668 lines), zero `#[test]`.

## Existing coverage (narrower than "none")

`tests/test_capture_worker_smoke.py` (3 tests) does exercise framing and the handshake, and it **does** run in CI via the `changes` job. The gap is the in-process Rust logic, not the protocol end to end.

## What is cheaply testable

All pure, all deterministic, all currently uncovered:

- `SpscRing` push/drain/producer_position — `production.rs:33-82`
- `parse_nonce` — `production.rs:102-111`
- `read_control` framing — `production.rs:392-401`, takes `&mut impl Read`, so a byte cursor is enough

## Two gotchas that would make the tests decorative

**1. CI would not run them.** `cargo test --lib` from `app/src-tauri` covers the root `ui` package only. The capture crate needs an explicit `cargo test -p murmur-capture-helper` step, or the tests exist and never execute — the same failure mode this repo already had with the `tests/` directory.

**2. Sequential tests cannot cover `SpscRing`'s real risk.** It carries `unsafe impl Send`/`Sync` with mixed Relaxed/Acquire ordering (`production.rs:40-41, 55-77`). Single-threaded unit tests prove the algebra, not the memory ordering. If we are serious about that struct, it wants `loom`.

## Out of scope

TCC attribution, sandbox/entitlement behaviour, and notarized-bundle behaviour are not unit-testable and should stay in manual release QA.

### LINKED ISSUE #479: CI: add dependency vulnerability scanning (cargo audit, npm audit)
## Problem

There is no dependency vulnerability scanning anywhere in the repo — no `cargo audit`, no `cargo-deny`, no `npm audit` in any workflow.

## Why this outranks style tooling here

The threat model is unusually dependency-sensitive:

- Signed helper binaries and a notarized, distributed `.app`
- An auto-updater that installs code on user machines
- A local LLM sidecar and a capture helper, each with their own dependency trees
- AI agents and external CLI coders (codex / grok / composer) regularly adding dependencies, none of whom carry this repo's threat model in context

A vulnerable transitive dependency reaching a signed, auto-updating binary is a materially worse outcome than any lint finding.

## Proposed fix

Add to CI:

- `cargo audit` (or `cargo-deny check advisories`) over `app/src-tauri`, covering the workspace members — the app crate, both sidecars, and both protocol crates
- `npm audit --audit-level=high` over `app/`

Start advisory-only to establish the baseline, then gate once it is clean. Consider `cargo-deny`'s licence and duplicate-version checks as a second phase.

### LINKED ISSUE #480: Settings: migrate the stale 50 ms auto-paste delay to the 0 ms default
## Problem

Every install created before 2026-08-04 still sleeps 50 ms before each paste, for a delay that was already removed.

Commit 6046f52 ("Reduce dictation latency floor") changed the default to `0`:

```ts
// app/src/lib/settings.ts:374-376
// Native CGEvents can paste immediately in the common case. Apps that move
// focus asynchronously can still opt into a settling delay in Settings.
autoPasteDelayMs: 0,
```

The Rust default matches (`state.rs:302`). But `loadSettings` (`app/src/lib/settings.ts:513-517`) only validates and clamps the persisted value — it never migrates a legacy `50` to the new default. Existing installs keep the old value indefinitely.

## Evidence

From the v0.28.0 dogfood install (see #472), n=202 injections:

| field | p50 | p95 | max |
|---|---|---|---|
| clipboard_ms | 0 | 0 | 2 |
| **delay_ms** | **50** | **50** | **50** |
| focus_ms | 0 | 24 | 1510 |
| key_event_ms | 3 | 4 | 15 |
| total_ms | 55 | 83 | 1565 |

All 202 injections logged `delay_ms=50` — never any other value. The sleep is **91% of median paste time**, and it is pure wall-clock: the actual guard, `focused_field_state()`, runs *after* it (`injector.rs:167` sleeps, `:176` checks). Across those 202 pastes there were zero first-attempt failures, zero retries, zero CGEvent failures, zero osascript fallbacks, and zero non-editable rejections.

## Proposed fix

A one-time migration in `loadSettings` that resets the exact legacy default `50` to `0`, gated on a persisted settings-version flag so it runs once.

## Two honest caveats

**The value is genuinely user-settable** via the paste-delay slider (`SettingsPanel.tsx:871`), so a deliberate user-chosen `50` is indistinguishable from a stale default. The version flag limits the clobber to a single occurrence rather than eliminating it; anyone who wanted 50 can set it again and it will stick.

**Nobody in the telemetry has actually run `delay=0`.** The zero-failure record above was measured *at* 50 ms. This migration is the soak test, not the confirmation of one — so it is worth watching paste failure/retry counts after it ships.

## Payoff

~45 ms off every dictation for affected installs — roughly 9% of the ~485 ms end-to-end pipeline, and the cheapest user-visible latency win currently identified.

### LINKED ISSUE #481: Repo: land the unpushed work, prune merged branches, decide the open PRs
## Problem

Completed and in-flight work is accumulating in places that are invisible to the repo and easy to lose.

## Unpushed work that exists in one place only

- **`stash@{0}`** — "pre-clean-main-2026-08-06 audio lifecycle and UI work": 24 tracked files (+360/−73) plus 7 untracked files, ~1,600 insertions total, including `scripts/murmur-audio-triage.sh` and the audio-lifecycle plan docs. On no branch, on one machine.
- **Branch `chore/ci-gates-and-log-noise`** (commit `7f7a0a9`) — CI integration-suite gating and the log-noise fix. Committed, no upstream.
- **Five stashes total**, three of them named with a `worktree-cleanup: preserve ...` pattern dating to 2026-07-22. The recurring name suggests this is habitual rather than incidental.

## Stale branches

31 of 38 remote branches are already merged into `main` and were never deleted. 6 are unmerged, two stale since late July (`cursor/theme-engine-plan`, `cursor/setup-dev-environment-b220`).

## Open PRs

- **#444 "Harden updater state ownership and release manifest policy"** — open since Aug 3. This is *updater-hardening* work; security work left to rot is worse than security work never started. Land it or close it deliberately.
- **#440** (docs: Cursor Cloud Linux setup notes) — draft since Aug 3
- **#431** (docs: post-release production latency comparison gate) — draft since Aug 2

## Proposed fix

1. Push `stash@{0}` to a real branch; triage the other four stashes and drop what is dead
2. Push `chore/ci-gates-and-log-noise` and open its PR
3. Delete the 31 merged remote branches; enable auto-delete-on-merge so this does not recur
4. Decide #444 explicitly — merge or close with a reason
5. Decide #440 and #431 — both are docs-only and cheap either way

Items 1 and 2 are the only ones with a loss risk; the rest is tidiness.

### LINKED ISSUE #482: CI: clear the clippy baseline and gate on it with a pinned toolchain
## Problem

No Rust lint gate exists. `cargo clippy` has never run in CI, and there is no `clippy.toml`.

Current baseline, `cargo clippy --all-targets`: **63 warnings**, all stylistic, 29 auto-fixable.

Breakdown:

| count | lint |
|---:|---|
| 7 | this `impl` can be derived |
| 5 | unneeded `return` statement |
| 5 | manual char comparison |
| 5 | `map_or` can be simplified |
| 4 | assertion has a constant value |
| 3 | very complex type |
| 3 | manual `.is_multiple_of()` |
| 3 | field assignment outside initializer |
| 2 | `map_err` over `inspect_err` |
| 2 | too many arguments (8/7) |
| 2 | non-binding `let` on a future |
| 2 | manually constructing a nul-terminated string |
| 2 | large size difference between variants |

Nothing here is a latent bug — the historical defects in this repo came from CoreAudio, sandboxing, and TCC, none of which clippy can see. The value is hygiene on AI-authored code, and it should be priced that way rather than oversold.

## Landmine to avoid

`ci.yml:122` uses `dtolnay/rust-toolchain@stable`, which floats. Combining a floating toolchain with `-D warnings` means CI breaks on **every Rust release that introduces a new lint** — a false red on unrelated PRs, which at this repo's sub-hour merge cadence is exactly the failure mode that gets gates disabled.

## Proposed fix

1. Pin the toolchain (`rust-toolchain.toml` or an explicit version in the action)
2. Clear the 63-warning baseline — `cargo clippy --fix` handles 29 of them
3. Add `cargo clippy --all-targets -- -D warnings` to the `rust-macos` job

Note the `--all-targets` dependency on #478 — without it, `tests/` is not linted either.

### LINKED ISSUE #483: CI: adopt rustfmt with a one-shot format and .git-blame-ignore-revs
## Problem

No `rustfmt.toml`, and `cargo fmt --check` has never run in CI. Current baseline: **224 files** would change.

## Why this is worth doing despite the churn

The instinct is to skip it — one maintainer, no style disagreements, and a 224-file diff is ugly. That reasoning is wrong for this repo specifically: **grok, codex, composer, and Claude subagents all write Rust here.** None of them share a formatting habit, and a solo maintainer reviewing machine-authored diffs is the strongest case for canonical formatting, not the weakest. Formatting noise in an AI-generated diff is exactly what hides a real change during review.

## The real cost, and how to pay it

Not `git blame` — that is solved by committing the format as one isolated commit and adding it to `.git-blame-ignore-revs`, which GitHub honours natively.

The real cost is **conflicts with in-flight work**. Currently outstanding: `stash@{0}` (31 files), 3 open PRs, and several unmerged branches. A 224-file reformat landing first would force every one of them through a painful rebase.

## Proposed sequence

1. Land or close `stash@{0}` and the open PRs (#444, #440, #431) — see #481
2. Add `rustfmt.toml` (defaults are fine unless there is a house preference)
3. One isolated commit: `cargo fmt` across the tree, no other changes
4. Add that commit's SHA to `.git-blame-ignore-revs`
5. Add `cargo fmt --check` to the `rust-macos` job

Step 3 must contain nothing but formatting so the ignore-revs entry stays honest.

### LINKED ISSUE #484: Release: cut CHANGELOG version sections as part of the release process
## Problem

The CHANGELOG has not been cut into a version section since 0.23.9, while the app ships 0.28.0.

`CHANGELOG.md:7` is `## [Unreleased]`, and `CHANGELOG.md:76` is `## [0.23.9] - 2026-07-30`. Everything from 0.24 through 0.28 — five minor releases and several patches — sits undifferentiated in the `[Unreleased]` bucket. 18 commits have landed since the file was last touched.

## This is a process bug, not a content bug

Worth being precise: `[Unreleased]` is **actively and well maintained** — the entries are detailed and current. Nobody is failing to write changelog entries. What is missing is the release step that renames `[Unreleased]` to `[X.Y.Z] - DATE` and opens a fresh `[Unreleased]` above it.

So the fix belongs in the release process, not in a documentation backfill.

## Proposed fix

1. Add the section-cut to the release flow — the natural home is alongside the existing `chore: bump version` step, which already edits `tauri.conf.json` and `Cargo.toml`
2. Optionally have `release.yml` fail the release when the top section is still `[Unreleased]` with the tag already assigned, so this cannot silently drift again
3. Backfill 0.24 → 0.28 headers once, best-effort, from the git log

## Related discrepancy

`app/package.json` still reports `"version": "0.1.0"` while `tauri.conf.json` and `Cargo.toml` both report the real version. Harmless today, but it is a raw inconsistency and the release step could keep all three aligned.

### LINKED ISSUE #485: Docs: close the reference drift and check command count in CI
## Problem

The reference docs are the source of truth for both humans and AI agents working in this repo, and several have drifted from the code. One of these drifted **within 24 hours** during a review, which is the argument for checking it mechanically rather than by hand.

## Confirmed drift

**Command count.** `lib.rs` registers **111** commands in `generate_handler![...]`; `CLAUDE.md` says "110" in two places. `has_visible_windows` is registered but missing from `docs/reference/commands.md`. Everything else matches — 110 of 111 documented.

**Events.** Two events are emitted and consumed but undocumented in `docs/reference/events.md`:
- `recording-interrupted` — emitted `commands/recording.rs:2352`, listened for in `useRecordingState.ts:124` and `useOverlayRuntime.ts:170`
- `transform-capture-failed` — emitted `transform_flow.rs:1283`

**File Map.** `CLAUDE.md`'s File Map has no broken paths, but omits 15+ source files over 300 LOC, including `knowledge_store/repository.rs` (1,319), `performance_metrics/repository.rs` (1,046), `spoken_numbers.rs` (745), `spoken_structure.rs` (687), `capture_agent_probe.rs` (1,050), `capture_helper_probe.rs` (726), `hang_diagnostics.rs` (351), plus the entire `sidecars/capture/` and `crates/capture-helper-protocol/` family.

## Verified accurate (leave alone)

`docs/reference/settings.md` matches `settings.ts` exactly — 35 of 35 fields, both directions.

## Proposed fix

1. One opportunistic pass to correct the count, document `has_visible_windows` and the two events, and add the missing File Map rows
2. Add a CI check that compares the `generate_handler!` entry count against the row count in `docs/reference/commands.md` and fails on mismatch

Step 2 is the durable half — a hand-fix drifts again by the next feature.

georgenijo:
## Outcome

This is the single cumulative review PR for the sequenced health sweep. All issue PRs were tested independently and merged only into `codex/health-sweep-integration`. This PR is intentionally left unmerged for maintainer review; `main` has not been changed by the sweep.

## Health issues, in required order

1. **#476** — stop log shipping from spawning microphone helpers; retain privacy-safe cached audio state
2. **#477** — make `ci-pass` include visual regression and validate the aggregate gate
3. **#478** — add capture-worker framing, nonce, and SPSC ring unit coverage
4. **#479** — add advisory Rust and npm dependency audits without blocking healthy builds
5. **#480** — migrate legacy paste delay once while preserving explicit user settings
6. **#481** — protect in-flight work, land the CI prerequisite, prune only proven-merged branches, enable merged-head cleanup, and resolve the exact stale PR queue
7. **#482** — pin Rust 1.96.0, clear the Clippy baseline, and deny warnings in CI
8. **#483** — last by design: adopt canonical rustfmt, isolate the format commit, add its exact blame-ignore SHA, and enforce formatting in CI
9. **#484** — synchronize all release version surfaces and require changelog cuts through tested release tooling
10. **#485** — synchronize the Tauri reference contract at 112 commands and prevent future command/event drift in CI

## Stacked PR receipt

- prerequisite and #481 foundation: #488
- #476: #493
- #477: #494
- #478: #496
- #479: #497
- #480: #498
- #482: #499
- current-main synchronization: #501
- #484: #502
- #485: #503
- #481 preserved updater decision: #444
- #481 temporary release-latency bridge: #431
- #483 final formatting baseline: #504

Stale PR #440 was closed with a recorded rationale. The preservation branch, all five stashes, and out-of-scope open PRs #489, #490, #491, #492, and #495 were left intact.

## Cleanup receipt

- 43 exact remote branches were deleted only after ancestry proof
- `delete_branch_on_merge` is enabled for future merged heads
- `issue/481-preserve-audio-lifecycle-work` remains preserved
- no unrelated repository setting was changed
- the durable production-latency feature remains tracked by open issue #430

## Validation

Every stacked PR passed its applicable GitHub matrix before integration. On the final tree:

- `cargo fmt --all -- --check`
- `cargo clippy --all-targets -- -D warnings`
- `cargo check --all-targets`
- Rust: 750 passed, 5 ignored, plus all integration suites
- capture helper: 6 passed
- TypeScript: `npx tsc --noEmit`
- frontend: 71 files, 661 tests passed
- Python policy suite: 114 tests passed
- workflow policy: 5 CI cases, 4 release-build cases, 3 secretless rehearsal jobs, 13 publication gates
- reference drift check: 112 registered commands and 112 documented rows
- release version check: all surfaces and newest changelog agree on 0.28.1
- `git diff --check origin/main...codex/health-sweep-integration`
- final integration tree equals the fully green #504 head
- integration is 0 commits behind and 33 commits ahead of current `origin/main`

Closes #439.
Closes #476.
Closes #477.
Closes #478.
Closes #479.
Closes #480.
Closes #481.
Closes #482.
Closes #483.
Closes #484.
Closes #485.

## Summary by CodeRabbit

* **New Features**
  * Added a clear “Preparing update…” state before downloads begin.
  * Prevented overlapping update checks and installations.
  * Added configurable minimum-version update policies with clearer failure reporting.
  * Preset matching now ignores surrounding punctuation.
  * Added overlay vertical-position adjustment support.

* **Bug Fixes**
  * Migrated legacy auto-paste settings to the current zero-delay behavior while preserving intentional settings.
  * Reduced unnecessary microphone enumeration during diagnostics and log shipping.

* **Documentation**
  * Updated updater, release, event, command, and architecture documentation.

* **Chores**
  * Standardized release version validation and pinned build tooling for consistent releases.


georgenijo:

PR #506, opened 2026-08-07
chore: consolidate health, reliability, and latency work

### LINKED ISSUE #340: Transform: reduce end-to-end latency (warm-on-arm, skip AX retry on browsers, token streaming)

### LINKED ISSUE #384: Audio: dictation status lingers at Processing after a timed-out Stopping teardown

### LINKED ISSUE #388: Audio: blocked recovery is a dead end — stuck stream_build leaves app in Recovering until restart (and restart doesn't help)
## Summary

When the Core Audio HAL call inside the `stream_build` phase (cpal `build_input_stream`) never returns, the #380 audio lifecycle supervisor enters **Recovering and can never leave it**. Recovery only exits when the worker thread joins (`finish_attempt` in `audio_lifecycle.rs` — triggered by `ThreadExited`), the worker is stuck in an uncancellable FFI call, and every new start is rejected with `AudioRecovering`. The app stays wedged on "Recovering microphone…" until the app is restarted — and on a machine whose audio stack is wedged system-wide, the next attempt hangs identically, so the in-app guidance ("Restarting Murmur clears this attempt") never actually gets the user back to a working state.

Observed in the field on a MacBook Air running 0.22.1 (auto-updated from 0.21.3 mid-log). Not a regression introduced by #380: the same machine showed the same underlying HAL failure on 0.21.3 (5s init timeouts, a 14s "recording" that captured 0 samples, 9.6s teardown, and a stranded stream build that returned ~6.5 minutes later with "device no longer available"). #380 made the failure visible and contained — but also terminal.

## Evidence (user log, redacted)

**On 0.21.3 (before update) — the HAL was already sick:**

```
19:06:50.046 start_native_recording: device=MacBook Air Microphone recording_id=1
19:06:55.048 ERROR start_native_recording: audio failed: Audio thread failed to initialize within timeout

19:11:58.338 start_native_recording: device=MacBook Air Microphone recording_id=2
19:12:01.684 run_audio_capture ...                      <- stream build took 3.3s
19:12:15.764 stop_recording: raw_samples=0, wall_secs=14.1, audio_secs=0.0
19:12:15.764 audio teardown + resample: 9.634583916s    <- 0 samples in 14s, 9.6s teardown

19:12:19.491 start_native_recording: recording_id=3
19:12:24.493 ERROR ... failed to initialize within timeout
19:12:29.988 run_audio_capture ...                      <- stranded build appeared 10.5s in
19:19:06.389 ERROR Audio capture error: Failed to build stream: The requested device is
             no longer available.                       <- ~6.5 MINUTES after the attempt
```

**On 0.22.1 (after update) — every attempt hangs in `stream_build`, recovery never exits:**

```
13:34:57.353 audio initialization phase entered  {phase:"device_enumeration"}   (106ms, fine)
13:34:57.459 audio initialization phase entered  {phase:"config_lookup"}        (12ms, fine)
13:34:57.472 audio initialization phase entered  {phase:"stream_build"}         <- never exits
13:35:02.357 WARN audio initialization still connecting {elapsed_ms:5005}
13:35:27.354 audio attempt entering recovery {cancellation_reason:"hard_deadline"}
13:35:27.354 ERROR audio initialization failed {"Microphone initialization exceeded the 30 second deadline"}
13:35:37.354 WARN audio recovery remains blocked
             ... (Recovering forever; all subsequent starts rejected until app restart)
```

The user restarted the app 4 times across the log; each fresh launch reaches `stream_build`, hangs, and re-enters the same terminal Recovering state (some via the 30s hard deadline, some via user cancellation — same outcome).

Corroborating system-wide wedge: one launch's Core ML model load took **30,967ms** (`model_prepare_complete load_ms=30967`) vs ~100–200ms on every other launch in the same log.

## Why it's stuck (code path)

- `stream_build` runs cpal `build_input_stream` — a blocking, uncancellable Core Audio HAL call.
- Cancellation (user or hard-deadline) moves the attempt to `AttemptPhase::Recovering`.
- Recovering exits only via `AudioWorkerEvent::ThreadExited` → `finish_attempt` → `handle.join()`.
- While the attempt exists, `start` returns `AudioStartError::AudioRecovering` — by design (#380: no overlapping owners).
- If the HAL call never returns, the worker never exits, recovery never finishes, and the pipeline is bricked until restart.

## Proposed direction

1. **Bounded orphan-after-deadline (preferred):** after N seconds of blocked recovery (e.g. 60s), mark the generation dead, leak the stuck worker thread, and return to Idle so a fresh attempt is possible. The #380 machinery already suppresses late readiness and late samples from cancelled generations, and a Stop command is queued for the zombie; the residual risk is a transiently duplicated HAL stream if the zombie eventually wakes. This is the pre-#380 detach re-admitted — but explicit, bounded, logged, and once per generation.
2. **Escalated wedged-audio guidance:** when blocked recovery persists (or recurs across launches), tell the truth the current banner can't: "macOS audio appears stuck system-wide — check other audio software (virtual/aggregate devices), or reboot." Restarting Murmur alone does not help in this state.
3. Do both: orphaning restores in-app recoverability; guidance covers the case where the whole HAL is wedged and any retry will hang too.

## Related

- #380 — introduced the single-owner supervisor and the Recovering state (this issue is a designed-in corner of it, not a regression).
- #384 — status lingers at Processing after a timed-out Stopping teardown. Same family: uncancellable Core Audio call vs. single-owner join. Consider resolving both with one design decision.

### LINKED ISSUE #439: Harden updater state ownership and release-manifest policy

### LINKED ISSUE #446: Audio: surface an on-device capture-health signal when the primary backend chronically fails over
## Context

The v0.25.0 latency incident (fixed by #445) ran unnoticed across four installed versions: AUHAL hung in `AudioOutputUnitStart` on every recording, the bounded fallback made recordings *succeed* after ~8.4s, and nothing watched the numbers. The app had the evidence by recording #2 of the first affected session — `startup_ms`, winning backend, and fallback occurrence are all recorded per run in `performance_metrics/`.

## Proposal

Add a small judgment layer on top of existing measurements, surfaced through the existing health plumbing (`usePerformanceHealth`, Settings → Performance):

- Flag when the last N recordings (e.g. 5) all required primary-backend fallback before first PCM, or when rolling p50 `startup_ms` exceeds a threshold (~2s).
- The #445 session memo is a free signal: "memo promoted the fallback backend at first PCM" recurring across many app launches = chronic primary-backend failure worth telling the user about.
- Presentation: a health line in the Performance page (and optionally a one-time notice), naming the degraded backend and the observed startup cost. No restart advice (a coreaudiod wedge survives restarts — see docs/draft incident notes).

## Constraints

- Pure read-side aggregation over data already collected; no new capture-path behavior.
- Privacy: thresholds and backend names only — no device labels/UIDs, matching existing telemetry rules.

Related: #430 (on-device cross-version latency comparison), #445 (session backend memo).

### LINKED ISSUE #447: Telemetry: scheduled regression watch over shipped capture-startup metrics
## Context

The v0.25.0 incident (#445) was diagnosed by hand-grepping one shipped `events.jsonl`. The log shipper already centralizes these files; what's missing is anything that reads them on a schedule. v0.24.2 shipped an install with *zero* ready recordings and no one knew until a debugging session days later.

## Proposal

A scheduled job (cron routine / fleet agent) on the log-receiver side that periodically aggregates, per install and per app version:

- `startup_ms` percentiles from `audio readiness accepted`
- counts of `capture backend exceeded its active initialization budget`, split by `backend` and (post-#445) `last_setup_step` — the step histogram names the hanging native Core Audio call directly
- fallback frequency (`trying bounded fallback`) and both-backends-failed events
- ready-recordings-per-session (zero across a session = the v0.24.2 total-failure signature)

Alert on version-over-version regression (e.g. a new version's p50 startup > 2× the prior version's for the same install) and on any install with repeated zero-ready sessions.

## Constraints

- Shipped logs are already privacy-stripped; the watch adds aggregation only, no new collected fields.

Related: #434 (persist shipped logs in SQLite + dashboard search — that store would be the natural query substrate for this watch), #430, #445.

### LINKED ISSUE #476: Audio: stop spawning a capture helper every 60 s from the log shipper

### LINKED ISSUE #477: CI: make ci-pass an honest required check

### LINKED ISSUE #478: Capture worker: unit-test the pure framing and ring-buffer logic

### LINKED ISSUE #479: CI: add dependency vulnerability scanning (cargo audit, npm audit)

### LINKED ISSUE #480: Settings: migrate the stale 50 ms auto-paste delay to the 0 ms default

### LINKED ISSUE #481: Repo: land the unpushed work, prune merged branches, decide the open PRs

### LINKED ISSUE #482: CI: clear the clippy baseline and gate on it with a pinned toolchain

### LINKED ISSUE #483: CI: adopt rustfmt with a one-shot format and .git-blame-ignore-revs

### LINKED ISSUE #484: Release: cut CHANGELOG version sections as part of the release process

### LINKED ISSUE #485: Docs: close the reference drift and check command count in CI

georgenijo:
## Outcome

This is the single cumulative replacement for #505, #489, #490, #491, #492, and #495. It combines the completed codebase health sweep with the reliability and transform-latency queue on top of the sweep's final repo-wide rustfmt commit.

## Included work

- Health sweep: CI/runtime hygiene, capture-helper hardening, paste-delay migration, rustfmt, and strict Clippy coverage (#439, #476–#485).
- Reliability: recover from stuck stream construction (#388), clear Processing after late teardown (#384), surface chronic capture failover on device (#446), and schedule capture-startup regression watches (#447).
- Speed: warm transform on arm, skip browser AX retry, and stream rewrite tokens into the review experience (#340).
- Integration-only reconciliation: rustfmt of the post-sweep audio cherry-pick plus two Rust 1.96 lint fixes.

## Validation

- `cargo fmt --all -- --check`
- `cargo clippy --all-targets -- -D warnings`
- `cargo check --all-targets`
- `cargo test -- --test-threads=1`
- `npx tsc --noEmit`
- `npm test` — 72 files, 672 tests
- `python3 -m unittest discover -s tests` — 126 tests

Closes #439
Closes #476
Closes #477
Closes #478
Closes #479
Closes #480
Closes #481
Closes #482
Closes #483
Closes #484
Closes #485
Closes #388
Closes #384
Closes #446
Closes #447
Closes #340

Reliability umbrella: #486
Speed umbrella: #487


georgenijo:
Integration receipt: Fleet-synced the clean MacBook checkout to head 223f3e5, rebuilt npm dependencies and all bundled helpers, and launched the detached Tauri dev app. Vite is listening on localhost:1420 and target/debug/ui is running. Full GitHub CI is green.

georgenijo:
MacBook dev-launch correction: direct Fleet/SSH process ancestry caused macOS TCC to attribute the capture worker to sshd and deny microphone access after the first PCM frames. Relaunched the same clean head through the logged-in Terminal.app session, initiated via Fleet. TCC now attributes the worker to com.apple.Terminal and returns authValue=2 (allowed); Vite and target/debug/ui are healthy. No source delta was required.

georgenijo:
Notch overlay follow-up: fixed the macOS 26 Stage Manager failure where the layer-25 overlay remained attached to Murmur’s previous app set. The native overlay now opts into CanJoinAllApplications in addition to all-Spaces, full-screen auxiliary, stationary, and cycle-exclusion behavior. Verified on the fleet MacBook at 9b5a4ec: while another Stage Manager set was active, WindowServer reported the 257×33 overlay at (627,0), layer 25, kCGWindowIsOnscreen=1. Full local validation passed: cargo fmt/check/test/clippy, TypeScript, and 72 Vitest files / 672 tests.

PR #527, opened 2026-08-11
CI: smoke capture worker to first PCM

### LINKED ISSUE #448: CI: native capture-worker smoke to first PCM on a self-hosted mac runner
## Context

The scariest regression class in the capture path is one that breaks AUHAL on healthy machines and hides behind the CPAL fallback — recordings still succeed, unit tests still pass, and only startup latency changes. The bundled `scripts/smoke_test_capture_worker.py` stops at the `streamOpen` phase, before any real Core Audio work. During #445 review a deeper smoke was written (session scratchpad) that drives the worker through a full `start` on each backend to first PCM, asserting per-native-call setup-step order and balance.

## Proposal

- Promote the deep smoke into `scripts/` (extend or replace `smoke_test_capture_worker.py`): hello → start(auhal) → assert setup steps `device_resolution … stream_start` entered+completed in order → ≥3 PCM frames → stop/stopped; repeat for cpal. Assert a startup-time bound (e.g. first PCM < 2s) so latency regressions fail loudly.
- Run it as a CI job on a self-hosted macOS runner (the mac mini) on capture-path changes (`sidecars/capture/**`, `crates/capture-helper-protocol/**`, `src/audio*.rs`).
- **Prerequisite:** a stable audio input on the runner. The mini's only input today is a Continuity iPhone mic that goes zombie (listed by CoreAudio, fails `default_input_config`/`set_input_callback`) — observed 2026-08-04, identically on pre- and post-#445 workers. Install a virtual device (e.g. BlackHole 2ch) and target it explicitly by UID so the smoke is deterministic.
- TCC note: the runner context needs a persisted microphone grant (or the virtual device, which avoids TCC entirely — verify).

## Constraints

- Smoke must keep the worker's bounded/killable contract: hard timeout + kill on hang, nonzero exit.
- No PCM content leaves the runner; assert on counts/timings only.

Related: #445, #436.

georgenijo:
## Summary
- extend the production-v3 worker smoke with explicit-device AUHAL and CPAL runs through first PCM, three PCM frames, and graceful stop
- assert exact native setup-step ordering, a two-second first-PCM bound, frame sequencing, and hard process-group termination without decoding or exporting PCM
- gate capture-path changes on a labeled self-hosted Apple Silicon macOS audio runner and protect the workflow contract with mutation tests

## Test plan
- [x] `cd app/src-tauri && cargo check`
- [x] `cd app/src-tauri && cargo test -- --test-threads=1`
- [x] `cd app && npx tsc --noEmit`
- [x] `cd app && npm test -- --run` (697 tests)
- [x] Python workflow/smoke/policy suite (123 tests)
- [x] real capture-worker protocol-only smoke
- [ ] native AUHAL + CPAL first-PCM smoke — awaiting runner prerequisites below
- [x] Murmur Bench: N/A — this PR changes only CI/smoke infrastructure and does not change recognition, delivered text, model runtime, VAD, or production capture behavior

## Runner prerequisite receipt

As of 2026-08-11, the repository reports zero registered self-hosted runners and has no `MURMUR_CAPTURE_DEVICE_UID` repository variable. The current Mac mini exposes no audio input in `system_profiler`, has no BlackHole installation, and the worker fails closed with `noInputDevice` / `retainedSamples: 0` for an explicit unavailable UID.

To make this check green, register the Mac with the standard `self-hosted`, `macOS`, and `ARM64` labels plus `murmur-audio`; install/configure a stable virtual input; and set `MURMUR_CAPTURE_DEVICE_UID` to its Core Audio UID. The job then verifies both AUHAL and CPAL against that exact input without retaining or uploading PCM content.

Closes #448

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,7 @@ jobs:
       frontend: ${{ steps.filter.outputs.frontend }}
       rust: ${{ steps.filter.outputs.rust }}
       github: ${{ steps.filter.outputs.github }}
+      capture: ${{ steps.filter.outputs.capture }}
     steps:
       - uses: actions/checkout@v4
 
@@ -63,6 +64,13 @@ jobs:
               - 'tests/test_capture_agent_matrix.py'
               - 'tests/test_capture_worker_smoke.py'
               - 'tests/test_log_receiver.py'
+            capture:
+              - 'app/src-tauri/sidecars/capture/**'
+              - 'app/src-tauri/crates/capture-helper-protocol/**'
+              - 'app/src-tauri/src/audio*.rs'
+              - 'scripts/smoke_test_capture_worker.py'
+              - 'tests/test_capture_worker_smoke.py'
+              - '.github/workflows/ci.yml'
 
       - name: Validate workflow policy
         run: |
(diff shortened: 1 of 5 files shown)


georgenijo:
Closing for now because the dedicated `murmur-audio` self-hosted macOS runner and stable `MURMUR_CAPTURE_DEVICE_UID` are not provisioned, leaving the required native capture check permanently queued. We can revisit the first-PCM gate when that infrastructure has an owner and is ready to operate.

PR #543, opened 2026-08-11
Integrate capture reliability, durable history, and release safeguards

### LINKED ISSUE #511: capture-watch: alert lifecycle bugs — zero-ready never clears, p50 baseline self-masks
From the post-merge fable review of #506 (reliability group, F1/F2/F3/F4/F6).

## Problems in infra/log-receiver/murmur-capture-watch.py

1. **Zero-ready alert never clears (medium, conf 8).** `zero_ready_sessions` accumulates over the entire retained append-only log per (install, version) with no windowing/decay. An install with 2 bad sessions in week one alerts hourly forever — permanent failed systemd unit, stale entries occupying the 20-alert cap — despite 200 healthy sessions since. The README's "a healthy later run clears the failed state" is false for this alert kind.
2. **Persistent p50 regression self-masks after one release (low, conf 9).** Baseline is always the immediately-preceding eligible cohort: v1 fast → v2 slow alerts, but once v3 (equally slow) accrues 5 samples the baseline becomes v2 and the alert silently disappears with the regression unfixed. Documented contract, but it is exactly the silent-failure mode the watch exists to prevent.
3. **Unbounded histogram (low, token-gated, conf 7).** `session_ready_histogram` keys are distinct ready-counts; a token-holding install can grow watch memory to the 256M MemoryMax and bloat capture-watch.json, which the receiver json.loads unbounded per dashboard render.
4. **`last_setup_step: "none"` collapses to "unknown" (cosmetic, conf 9).** audio.rs emits "none" when a backend timed out before any setup step; the watch merges it with unrecognized labels, losing the "hung before first native call" signal. One-line fix: add "none" to SETUP_STEPS.
5. **Cohorts mix transform captures (info, conf 9).** The watch counts any audio.capture_ready with no owner_kind filter; the on-device frontend (captureHealth.ts) is dictation-only. The two "capture health" signals are defined over different populations.

## Fix direction

Window zero-ready detection (e.g. last N sessions or last M days), consider anchoring p50 baseline to the best recent healthy cohort rather than strictly the previous version, cap histogram keys, add "none" to SETUP_STEPS, and pick one population definition for both signals.

## Landmines

- tests/test_capture_regression_watch.py pins current behavior — update fixtures deliberately, not incidentally.
- The report write is fsync+atomic and exit-2-after-write; keep that ordering.

### LINKED ISSUE #512: release: publish-time notes comparison is asymmetric and can false-fail
From the post-merge fable review of #506 (health group, F3, confidence 8).

## Problem

`release.yml` compares `gh release view --jq .body` (shell `$(...)` strips only trailing newlines) against the manifest `.notes`, which `scripts/release_artifacts.py` produced via Python `.strip()` (strips ALL leading/trailing whitespace including \r and spaces). Any GitHub web-UI save of the draft — even content-identical — can introduce CRLF endings or trailing spaces, making the equality check fail and blocking publish. Fail-closed only (no false-pass direction), but it will burn a release run.

## Fix

Normalize both sides identically before comparing — e.g. run both through the same Python `.strip()` (or normalize line endings + strip in jq). Add a case to the release tests with a CRLF/trailing-space draft body.

## Landmine

Do not weaken the check to a substring/fuzzy match — exact equality after shared normalization is the point (it guards against publishing a draft whose notes were edited away from the manifest).

### LINKED ISSUE #513: CI hygiene: rustfmt.toml ungated by paths filter; dependency-audit runs on bump commits; first-match step lookup
From the post-merge fable review of #506 (health group, F4/F6, confidence 8).

## Problems in .github/workflows/ci.yml

1. **rustfmt.toml matches no paths filter.** The `rust` filter covers `app/src-tauri/**` and root Cargo files; repo-root `rustfmt.toml` matches nothing, so a PR editing only formatting config runs no rust-macos job (ci-pass green via "skipped") and the divergence lands ungated, breaking the next unrelated rust PR at the fmt gate. Latent today (rustfmt.toml is comment-only), real the day someone sets an option.
2. **dependency-audit has no changes/bump-commit guard**, so it runs on `chore: bump version` pushes, and its setup steps (checkout, setup-node) lack continue-on-error, so infra flakes there put a red X on commits (ci-pass correctly ignores it — advisory by design, #479).
3. **validate_workflow_policy.py `named_step_block(ci, "Install Rust")` matches the FIRST of two "Install Rust" steps** — correct today only because rust-macos precedes dependency-audit in the file. Reordering jobs silently validates the wrong step.

## Fixes

Add `rustfmt.toml` to the rust filter; give dependency-audit the same bump-commit guard as `changes`; scope named_step_block lookups per-job (or assert uniqueness).

### LINKED ISSUE #514: diagnostics: capture-health panel starves after idle — ring buffer evicts capture events
From the post-merge fable review of #506 (reliability group, F5, confidence 7).

## Problem

`deriveCaptureHealth` (app/src/lib/captureHealth.ts) reads the 500-event telemetry ring (telemetry.rs), which also carries ~1/min system heartbeats plus shipper/system chatter. ~8h of idle (overnight) evicts all prior capture events, so the Performance tab regresses to "collecting data" most mornings even for heavy users — the "five-recording signal" is really "five recordings within the ring's survival window". Degrades safely (never a wrong verdict), just chronically uninformative. Secondary: usePerformanceHealth polls the full history every 2s while the tab is open.

## Fix direction

Either derive from a capture-scoped query (filter server-side / dedicated bounded buffer for capture events) or persist the last N capture outcomes independently of the general ring. Keep the insufficientData behavior — a wrong verdict is worse than a modest one.

## Landmine

The chronic-failover verdict (5/5 recovered fallbacks) was verified correct against owner-ID collision and windowing edge cases — don't loosen its thresholds while fixing data supply.

### LINKED ISSUE #521: History and stats are still localStorage-only — same durability gap settings had
The durable settings store (feat/durable-settings-store) moved the settings blob to a Rust-owned `settings.json` because WKWebView localStorage is lost on manual reinstall / WebKit eviction (observed in practice 2026-08).

Transcript **history** (`useHistoryManagement` / `lib/history.ts`) and **usage stats** (`lib/stats.ts`) still live only in localStorage and are exposed to exactly the same loss.

Options to evaluate:
- Reuse the opaque-blob pattern (a `history.json` via the settings-store module, generalized)
- Fold history into the existing SQLite knowledge/performance-store pattern (better fit for growing, queryable data with rolling trim)
- Decide whether stats are worth preserving at all or are acceptable to lose

Considerations: history is much larger than settings (rolling trim exists), contains transcript content (privacy: stays local either way), and `retainHistory=false` users should keep getting the discard-at-boundary behavior.

georgenijo:
## Summary

Integrates the five independently reviewed and validated changes from #526, #528, #529, #531, and #532 on top of current main:

- fix capture-watch alert lifecycle and baseline masking
- persist transcript history and usage statistics in Rust-owned durable storage
- normalize release-note comparisons symmetrically
- harden CI path filtering and workflow-policy validation
- retain privacy-safe capture-health observations across idle periods

Also preserves the Voice Query feature merged in #541 and resolves the combined registered-command documentation at 129 commands.

## Validation

- Individual PR CI: green
- Combined reference-doc validator: 129/129 commands, green
- Combined reference-doc tests: green
- Combined diff check: green
- Full combined CI: pending on this PR
- Murmur Bench: prior SHA passed controlled mirrored quick comparisons; current refreshed SHA requires rerun because main advanced

Closes #511
Closes #512
Closes #513
Closes #514
Closes #521

Supersedes #526, #528, #529, #531, and #532.

## Summary by CodeRabbit

* **New Features**
  * Transcript history and usage statistics now persist reliably across app restarts, with migration from existing cached data.
  * Added bounded capture-startup health diagnostics to help identify slow or fallback microphone initialization.
  * Performance diagnostics can now be cleared along with capture-health observations.
* **Bug Fixes**
  * History retention limits are consistently enforced when loading stored data.
  * Improved release-note verification handles line-ending and surrounding-whitespace differences consistently.
* **Documentation**
  * Updated command, settings, history, performance, and release documentation to reflect these changes.


georgenijo:
Final validation receipt for head b90163522d86a21108f73dfe46ce1ca96c171190 (base 0d0c927637082a2e52b5b943c85fe04e538f4076):

- cargo check: PASS
- Rust: 789 passed, 5 ignored; capture/LLM/transcription/transform integrations PASS
- TypeScript: PASS
- Frontend: 81 files / 718 tests PASS
- Python policy suite: 137 tests PASS
- diff check: PASS
- Murmur Bench gate: N/A — the PR delta contains post-capture diagnostics, durable post-delivery history/stats persistence, release tooling, and CI policy only. It does not change VAD, transcribers, model runtime, transcript transforms, benchmark execution, benchmarked dependencies, recognition output, or delivered text. Extra quick experiments were accuracy-neutral across all seven models but latency/memory results flipped with order and compilation heat, confirming they are non-gating harness variance rather than a causal regression in this diff. Raw reports remain local.

This head includes current main, including #544.

PR #576, opened 2026-08-13
feat(query): provider presets, auth preflight, actionable errors, and declared env vars

### LINKED ISSUE #550: feat(query): provider presets, auth preflight, actionable errors, and declared env vars
From the 2026-08-12 architecture review (the Claude "Not logged in" incident). The generic bridge stays; around it:

- **Presets** (Claude / Codex / Grok / Cursor / Custom) as data: binary discovery paths, recommended args, auth-probe command, auth-failure signatures, permitted env additions (`CLAUDE_CONFIG_DIR`, `CODEX_HOME`).
- **Preflight**: run `validate_command` at enable time (not first keypress) + a Settings "Test" button that runs the preset's auth probe (`claude auth status`, …) through the identical `spawn_user_cli` path; bounded stdout+stderr shown in Settings UI only, never telemetry.
- **Bounded stderr capture** (~16 KiB tail) on the query path; on failure show as error detail distinct from answer content; map known auth signatures to `provider_not_authenticated` with the exact fix ("Run claude /login in Terminal") and a one-click "Sign in…" button that launches the vendor CLI's own login (Terminal-launch tier 1; direct spawn tier 2 per-provider once proven headless-safe) and auto-re-probes until green.
- **Declared env vars** (review Option G): explicit name+value pairs on top of the fail-closed allowlist; deny-list `HOME` and allowlist-key overrides; persist Rust-side, not localStorage; sensitive-value storage design before accepting secrets.
- Fix precedence bug: `QueryReviewApp.tsx` renders `driver.answer || errorMessage` — stdout beats the error message on terminal failure.
- Add DECISIONS.md entries (allowlist rationale, stderr policy).

Murmur Bench: N/A — no benchmarked path.

georgenijo:
Closes #550.

From the 2026-08-12 architecture review (the Claude "Not logged in" incident). The generic bridge is unchanged — one absolute executable, fixed argv, no shell — this is the data and diagnosis around it.

## What changed

**Provider presets** (`query_presets.rs`) — static data for Claude Code, Codex, Grok, and Cursor Agent: binary name, install-location fallbacks, one-shot arguments, auth-probe argv, vendor-login argv, "not signed in" signatures, and the env names each actually reads. Choosing a preset only *fills in* the executable and arguments; Murmur still spawns exactly what those fields say. Discovery probes the standard absolute prefixes, per-user prefixes under `$HOME`, each preset's private locations, then the host `PATH` — the fixed prefixes are what make it work for a Finder/Dock launch.

**Preflight and sign-in** — `validate_query_command` runs at enable time, so a missing path is refused during configuration instead of mid-question. Settings' "Test sign-in" runs the preset's probe through the identical `spawn_user_cli` path (same cleared environment, same declared pairs, same process-group ownership). Verdict is signature-first, exit-code-second, because `claude auth status` reports `"loggedIn": false` and still exits 0; an unexplained non-zero exit stays `unknown` and shows the raw output rather than guessing. "Sign in…" opens the vendor CLI's own login in Terminal and re-probes until green — Murmur never sees, prompts for, or proxies the credential.

**Bounded stderr and actionable errors** — stderr is piped and drained continuously (a full pipe would otherwise block the child) into a 16 KiB *tail*, since the explanatory line is the last one. On failure it is stored on the session and served to the popover through the same requester-gated command as the answer — never broadcast, logged, or persisted. A recognised auth signature remaps `exit_nonzero`/`process_failed`/`empty_answer` to `provider_not_authenticated` with the exact fix; codes describing Murmur's own bounds (`timed_out`, `termination_unconfirmed`, …) are never reinterpreted.

**Precedence fix** — `QueryReviewApp` rendered `driver.answer || errorMessage`, so a provider that printed "Not logged in" on stdout and exited non-zero looked as though it had answered. The failure now leads, with any partial stdout labelled as evidence below it.

**Declared env vars** (`query_env.rs`, review Option G) — explicit name/value pairs applied *underneath* the fail-closed allowlist, so nothing declared can shadow `HOME` or any other allowlist key; the spawn ordering enforces that again even if validation were bypassed. Refuses allowlist names, `DYLD_*`/`LD_*`, malformed names, duplicates, >16 pairs, >4 KiB values. Persisted Rust-side in `query-env.json` (0600) behind main-window-gated commands rather than the settings blob, so they never reach localStorage, and re-validated on read. Values are plain text and Settings says secrets do not belong there; Keychain-backed secret storage is deferred.

**DECISIONS.md** — two entries: the allowlist/declared-variable rationale, and the stderr capture + requester-gating policy.

## Verification

- `cargo fmt --all -- --check`, `cargo clippy --workspace --exclude murmur-llm-sidecar --all-targets -- -D warnings`: clean (macOS).
- `cargo test --workspace --exclude murmur-llm-sidecar --lib -- --test-threads=1`: 851 passed, 0 failed. 32 tests cover the new modules, incl. that a declared `HOME` cannot shadow the inherited one, that a signed-out Claude is recognised despite exiting 0, that the login command line is quoted rather than interpolated, and that a tampered env file is refused.
- `npx tsc --noEmit`: clean. `npx vitest run`: 784 passed, 0 failed (18 new), incl. a render test that the failure — not the CLI's stdout — is what the popover shows.

Murmur Bench: N/A — no benchmarked path. Capture, ASR, transcript transforms, and delivered text are untouched.


georgenijo:
@cursoragent review — post findings only, don't push fixes

georgenijo:
All seven review findings are fixed in `6553c2e5` (+ two follow-up commits for scope/rustfmt). Thanks — several were real fail-open holes.

| # | Finding | Resolution |
|---|---|---|
| 1 (High) | Deny list contradicted its own rationale — `NODE_OPTIONS`/`PYTHONPATH`/`BASH_ENV` etc. still injected code into the Node-based presets, and `HTTPS_PROXY`/`SSLKEYLOGFILE` could redirect or key-log provider TLS | Expanded to 24 names across both families (loader/interpreter, proxy/CA) plus the existing prefixes, **matched case-insensitively** — `https_proxy` was a free bypass of the uppercase-only list. Mirrored in `queryProviders.ts`. Kept as a deny list rather than an allowlist so the feature stays general, as the issue specified. |
| 2 | stderr drain never joined; the tail could still be in flight when the failure was classified | RAII `StderrDrain` guard joins on every exit path of `run_cli`. Every path already confirms or kills the owned group first, so the join is bounded. |
| 3 | Bare `"unauthorized"` remapped unrelated failures | Narrowed to `401 unauthorized` / `http 401` / `status 401`; a permission or quota error keeps its own code. |
| 4 | Bare exit 0 read as authenticated — overclaimed against the Settings copy | Added per-preset `auth_success_signatures` from each CLI's real signed-in output. Exit 0 without that positive signal is now `Unknown`, which shows the raw output. A custom command (no declared wording) still uses the clean exit, since that is all there is. |
| 5 | Probe had no exclusivity vs an in-flight query | `is_pipeline_busy` mirrors the gate `start_query_capture` applies (query/dictation/file transcription/benchmark/transform/corpus). |
| 6 | `stopWatching` left `until` in the future | Deadline zeroed on stop and an in-flight tick bails before rescheduling. |
| 7 | Empty env list could not be saved, so a deleted pair kept applying | Save is always available; labelled "Save (none)" at zero rows. |

Found while re-reading, same class as #2: `run_probe` joined its reader threads **before** confirming the process group was empty, so a wrapper CLI leaving a descendant on stdout would hang the Test button forever with the 30 s timeout already spent. Now confirms/kills the group first, mirroring `run_cli`. Covered by `a_probe_returns_even_when_a_descendant_outlives_the_command`.

Verification on `3bba5fbc`: `cargo fmt --check` and `clippy -D warnings` clean on macOS; **856 Rust tests pass** (5 new); `tsc` clean; **785 frontend tests pass**. Docs and the DECISIONS entry were updated to match the tightened rules rather than left describing the old ones.

Murmur Bench: N/A — no benchmarked path.

PR #578, opened 2026-08-13
feat: show Voice Query words while listening

### LINKED ISSUE #561: Voice Query: live word-by-word transcription while speaking
## What

While a Voice Query is **listening**, show the recognized words in the popover as they are spoken. When the query is sent, the partial text disappears and the popover shows only the answer.

Requested by George after using the merged #556/#557 work: there is currently no feedback on what Murmur actually heard until the agent answers, so a misrecognition is only discoverable from a confusing answer — after an agent call has been spent.

## Why this is not a small change

Murmur transcribes in one batch **after** capture stops: `stop_query_recording` hands the complete buffer to a single `transcribe_query` pass. No partial decoding exists anywhere in the app. This is backlog item #129 (streaming transcription), scoped down to the one place it is tractable.

**Scoping it to Voice Query listening is what makes it feasible.** During that window the ASR backend is otherwise idle — no dictation, transform, benchmark, or corpus capture can hold it (`blocks_capture`, #556). General streaming dictation remains out of scope.

## Approach

Follow the proven pattern from #545's microphone monitor (`microphone_preview.rs`): bounded periodic analysis during live capture, one inference in flight, stale generations discarded, work off the capture thread.

1. During `listening`, snapshot accumulated samples on a timer (~700ms), in the shape of `snapshot_if_due()`.
2. Decode the snapshot on the existing backend; skip the tick if one is already in flight.
3. `emit_to("query-review", "query-partial", …)` — the targeted channel answer chunks already use. **Never** the `query-state-changed` broadcast, which is deliberately content-free.
4. Guard every continuation with `query_pass_id`; a stale partial must never render over a newer pass.
5. On finish: clear partials, run the existing authoritative pass unchanged, render the answer.

## Constraints

- **Privacy:** spoken content must not reach telemetry, history, stats, file output, or any broadcast event. Partials are targeted to the `query-review` webview only, exactly like answer chunks. `docs/features/voice-query.md` records this boundary.
- **The final pass stays authoritative.** Partials are display-only and never feed the CLI, matching the mic monitor's "post-recording full buffer remains authoritative" rule.
- **Bounded cost.** Each tick re-decodes the buffer so far, so cost grows with question length. Needs an explicit cap (stop partials past N seconds, or switch to a rolling window) rather than unbounded growth.
- **Backend-dependent.** Parakeet CoreML on the ANE is fast enough (~118ms for 10s of audio). whisper.cpp on Metal likely is not at this cadence — decide whether it gets a slower tick or no partials.

## Known behavior

Partial text will revise itself as context arrives (a word can change once the next few land). This is inherent to re-decoding a growing buffer and is normal for streaming ASR, but it should be a deliberate, documented choice rather than a surprise.

## Out of scope

- Retry / re-ask without re-recording — separate ticket.
- Streaming for general dictation (#129 proper).
- Persisting questions to History; the current privacy boundary keeps query content out of it.

## Acceptance

- Speaking during a query shows words in the popover as they are recognized.
- Sending clears the partial text; only the answer is shown.
- Cancel/Escape mid-speech leaves no partial text behind.
- No spoken content in `events.jsonl`, history, stats, or any broadcast event.
- A stale pass cannot render partials over a newer one.

georgenijo:
## Summary
- While Voice Query is listening, Core ML Parakeet re-decodes the growing capture buffer about every 700 ms (one inference in flight, 20 s cap) and shows the words in the review popover.
- Partials are display-only: targeted to `query-review` via `query-partial`, never broadcast, never stored, never sent to the CLI. The existing post-stop `transcribe_query` pass stays authoritative.
- Sending, cancelling, or hiding the popover clears the partial. Whisper and CPU Parakeet stay silent during listening so a slow decode cannot stall finish.
- Rebased onto Voice Query v0.33.0 so listening words still work after the provider/history/context rewrite.

## Test plan
- [x] `cargo test -- --test-threads=1` (macbook after merging `origin/main`; 917 lib tests plus integrations, including peek/partial/pass-id tests)
- [x] `npx tsc --noEmit`
- [x] `npx vitest run` (844 passed)
- [ ] Native smoke on a built `.app` with Core ML Parakeet: speak during a query and confirm words appear; tap to send and confirm they vanish before the answer; Escape mid-speech leaves nothing; Whisper has no partials
- [x] Murmur Bench: N/A — Voice Query UI/event path; no dictation/VAD/backend/transform change on the bench corpus

Closes #561

## Summary by CodeRabbit

* **New Features**
  * Added live “Heard so far” speech previews while Voice Query is listening.
  * Partial speech updates refresh periodically and clear when listening ends, is cancelled, or fails.
  * Improved Core ML Parakeet audio handling for live decoding.
  * Added provider-specific safeguards for Claude queries to reduce unsupported tool and file-access assumptions.

* **Bug Fixes**
  * Prevented stale partial transcripts from appearing in later query passes.
  * Preserved recording while generating live speech previews.

* **Documentation**
  * Updated Voice Query and event documentation with partial transcript behavior and lifecycle details.


georgenijo:
## Post-review fixes at `3286350`

Two native bugs found in tonight's dev smoke, root-caused and fixed:

1. **Live partials froze at 20 s** — `PartialTick::Capped` killed the listening ticker. Replaced with a trailing 20-second decode window: partials keep updating until stop/cancel/fail, per-tick cost stays bounded, telemetry stays content-free (`capped` outcome retired).
2. **Claude answered with a fabricated `ls` of `voice-query-workspace`** — reproduced with the exact production argv from an empty scratch dir: with `--tools ""`, Claude Code still frames itself as a coding agent and *invents* file listings of its (empty) cwd as assistant text. The workspace was never written to and no `voice-query.db`/`session.log` exists on disk — no data leak, pure model fabrication. Fix: Rust-owned pinned `--append-system-prompt` argv pair for the Claude provider, appended at validation time so it covers existing saved Settings; verified against claude 2.1.228 (fabrication repro now answers correctly, normal questions unaffected, `--safe-mode` does not strip the flag). Isolation, no-shell spawn, env allowlist, and auth probe are untouched.

**Gates (mac-mini worktree @ 3286350):** `cargo test -- --test-threads=1` all green · `npx tsc --noEmit` clean · `npm test` 844/844 passed.

**Murmur Bench: N/A** — listening peeks are display-only and provider argv changes cannot affect recognition latency, accuracy, delivered-text output, or memory on the benchmarked dictation path.

PR #579, opened 2026-08-14
feat(query): retain opted-in contextual history

georgenijo:
## Summary
- Retain every recognized Voice Query when **Keep Voice Query history on this Mac** is enabled, including results produced with app/selection context and raw structured-provider fallback output.
- Keep the setting off by default and immutable per pass; no recognized question still means no history record.
- Do not add a separate context field. Saved provider output can quote supplied context, so Settings and History explicitly disclose that boundary.
- Remove the obsolete review-popover “Not saved to history” notice and update product/docs decision records.

## Why
The previous context/raw-fallback gate made the opt-in History surface appear broken for ordinary queries using the Selection context level. The redesigned contract matches transcription history: local retention is controlled by the explicit user setting.

## Review fixes
- Removed the dead structured-raw-fallback accessor and vestigial assertions.
- Inlined the empty adapter pass-through wrappers.
- Corrected and test-pinned the Settings privacy disclosure.
- Merged the latest `main` while preserving the live-partials work from #578.

## Validation
- `cargo fmt --check`
- `cargo clippy --workspace --exclude murmur-llm-sidecar --all-targets -- -D warnings`
- `npx tsc --noEmit`
- `python3 scripts/validate_reference_docs.py` — 158/158 commands
- `npm test` — 842 tests passed
- `cargo test -- --test-threads=1` — 922 unit tests passed, 5 ignored; all integration tests passed
- Exact-head native MacBook build produced and launched `Local Dictation Dev.app`; packaging proceeded through all bundles and stopped only at the expected missing updater private key in the local validation environment
- GitHub CI green on exact head `fd34eaeca52741da080ed99c10ee53369abd2f8d`

Murmur Bench: N/A — this changes local query-content retention and UI disclosure only; recognition, VAD, model runtime, delivery, and benchmarked paths are unchanged.


app/src/components/settings/SettingsPanel.tsx
```diff
@@ -1493,7 +1493,7 @@ export const SettingsPanel = memo(function SettingsPanel({
 
               <SettingToggle
                 title="Keep Voice Query history on this Mac"
-                description="Off by default. When on, Murmur keeps up to 200 questions, answers, provider IDs, token counts, durations, and stable errors in a separate Rust-owned local store. A pass that actually shares app context stays display-only and is not saved, because a provider may quote that context in its answer. Turning history off affects new queries; existing entries remain until you delete them from History → Queries."
+                description="Off by default. When on, Murmur keeps up to 200 recognized questions, answers, provider IDs, token counts, durations, and stable errors in a separate Rust-owned local store. This includes queries that shared app context: context is not stored as a separate field, but a saved answer may quote it. Turning history off affects new queries; existing entries remain until you delete them from History → Queries."
```
georgenijo, line 1496: **Stale absolute privacy claim left in this section.** This toggle description was correctly updated to say a saved answer may quote context — but the section footer ~55 lines below (line 1551) still reads:

> "Context content never enters history, saved files, usage stats, diagnostics, logs, or telemetry."

That absolute claim is now false and directly contradicts this toggle in the same Settings section. The same qualifier ("never stored as a separate field, but a saved/retained answer may quote it") was added in `docs/features/voice-query.md`, `docs/reference/settings.md`, `docs/reference/commands.md`, and `docs/features/history-workspace.md` — only this footer was missed. Notably, this PR deleted the `SettingsPanel.test.tsx` assertion that pinned the old sentence but left the sentence itself in the UI, so nothing guards it now. Suggest updating the footer with the same qualifier used everywhere else.

app/src-tauri/src/query_flow.rs
```diff
@@ -2083,39 +2035,22 @@ fn accept_stdout(
...
     bytes: &[u8],
 ) -> Result<(), &'static str> {
-    let updates =
-        adapt_stdout_for_pass(&app.state::<crate::State>().query, pass_id, adapter, bytes)?;
+    let updates = adapt_stdout_for_pass(adapter, bytes)?;
     accept_answer_updates(app, pass_id, sequence, updates)?;
     Ok(())
 }
 
 fn adapt_stdout_for_pass(
```
georgenijo: **Cleanup: these are now empty pass-throughs with misleading names.** After removing the `query`/`pass_id` params and the `mark_structured_raw_fallback` side effect, `adapt_stdout_for_pass` is exactly `adapter.push_stdout(bytes)` and `finish_adapter_for_pass` is exactly `adapter.finish()`. The `_for_pass` suffix now references per-pass bookkeeping that no longer exists — in a file where similarly named helpers genuinely do per-pass work, that invites wrong assumptions (or re-attaching logic "since the wrapper exists"). Suggest inlining at the two production call sites (`accept_stdout` line 2038, `run_cli` line 2496) and the test call sites; both adapter methods are `pub(crate)`, so nothing breaks.

app/src-tauri/src/query_flow.rs
```diff
@@ -3133,32 +3064,23 @@ mod tests {
...
         let snapshot = query.claim_terminal(pass_id).unwrap();
-        assert!(snapshot.structured_raw_fallback);
         assert_eq!(snapshot.answer, format!("{user_frame}{malformed}"));
         assert!(snapshot.answer.contains(private_context));
-        assert!(!persist_query_history_snapshot(&history, &snapshot, None).unwrap());
-        assert_eq!(history.list(0, 10, None).unwrap().total, 0);
+        assert!(persist_query_history_snapshot(&history, &snapshot, None).unwrap());
+        assert_eq!(history.list(0, 10, None).unwrap().total, 1);
 
         let raw_adapter = VoiceQueryAdapter::new(QueryProviderId::Custom, 4096);
```
georgenijo: **Vestigial assertion — copy-paste debris from the pre-PR test.** `raw_adapter` is constructed solely for `assert!(!raw_adapter.used_structured_raw_fallback())` and never used again; the `raw_snapshot` below is a hand-built literal with no data flow from it. Pre-PR this assertion motivated the snapshot's `structured_raw_fallback: false` field, but that field and all production reads of the flag are gone, so it no longer verifies anything about the behavior under test (raw providers never setting the flag is already covered in `query_adapter.rs` tests). Suggest deleting these two lines and the twin at ~3255 — which also removes two of the dead method's only remaining callers (see the CI-blocking finding in the review body).

georgenijo (commented):
Automated review (Claude Code): 4 findings, most severe first. Three are inline below; the top one is in a file the diff doesn't touch, so it's here:

**1. CI-blocking: `used_structured_raw_fallback()` is now dead code and fails clippy** — `app/src-tauri/src/query_adapter.rs:104`

This PR removed the method's last production callers (the `mark_structured_raw_fallback` hooks in `adapt_stdout_for_pass`/`finish_adapter_for_pass`), leaving only `#[cfg(test)]` call sites (`query_adapter.rs:1159/1400/1572/1595/1609`, `query_flow.rs:3086/3255`). The `pub(crate)` method has no `#[allow(dead_code)]`, so the lib target trips `dead_code`, which `cargo clippy --workspace --exclude murmur-llm-sidecar --all-targets -- -D warnings` turns into a hard error. This is already happening: the `rust-macos` check on head `ad72664c` fails with `error: method used_structured_raw_fallback is never used`, taking `ci-pass` down with it.

Suggested fix: delete the method and its now-vestigial test assertions (the adapter behavior is still covered through the internal `used_raw_fallback` field), or mark it `#[cfg(test)]`. Note this pairs with the inline comment on the vestigial `raw_adapter` assertion — removing those assertions is part of this fix.

app/src/components/settings/SettingsPanel.tsx
georgenijo: Fixed in fd34eaec: the Settings footer now carries the same qualified privacy language, and the frontend test pins the new copy and rejects the stale absolute claim.

app/src-tauri/src/query_flow.rs
georgenijo: Fixed in fd34eaec: removed both empty  wrappers and inlined  /  at production and test call sites.

app/src-tauri/src/query_flow.rs
georgenijo: Fixed in fd34eaec: removed the vestigial raw-adapter assertions together with the now-dead fallback accessor and all remaining test-only callers.

app/src-tauri/src/query_flow.rs
georgenijo: Correction: fixed in fd34eaec by removing both empty adapt/finish-for-pass wrappers and calling the adapter push_stdout and finish methods directly at production and test call sites.

georgenijo:
@coderabbitai review

PR #582, opened 2026-08-14
Show manual-paste cue for every clipboard-only delivery

### LINKED ISSUE #571: Notch: show the manual-paste cue for every clipboard-only delivery
## Problem

Murmur shows a brief `⌘V` cue in the notch when auto-paste is skipped after focus changes, but other successful clipboard-only delivery paths do not consistently show the same cue. Users can miss that text is ready and waiting for manual paste.

## Goal

Show the existing transient `⌘V` manual-paste cue whenever Murmur successfully puts an output on the clipboard and will not automatically paste it.

## Proposed contract

- Trigger from a confirmed clipboard write followed by a clipboard-only delivery outcome, not from listening for the Command key.
- Cover normal dictation when auto-paste is disabled, skipped, or safely fails.
- Do not show the cue after a confirmed successful auto-paste.
- Keep the cue transient, non-interactive, and accessible with the existing “Text copied to clipboard. Paste manually.” label.
- Do not include transcript or clipboard content in the event, logs, telemetry, or UI.
- Preserve pass/generation ownership so stale delivery events cannot flash over newer work.

## Acceptance criteria

- [ ] Every successful clipboard-only dictation delivery produces one bounded `⌘V` notch cue.
- [ ] Successful automatic paste does not produce the cue.
- [ ] Failed clipboard writes do not claim text is ready.
- [ ] Duplicate/stale events are ignored.
- [ ] Overlay tests cover disabled, skipped, failed, successful auto-paste, and stale-pass cases.
- [ ] Documentation explains that `⌘V` means text is waiting on the clipboard.
- [ ] Murmur Bench: N/A expected — presentation and delivery-status signaling only; run the gate if implementation unexpectedly changes delivered text or benchmarked paths.

georgenijo:
## Summary

- classify clipboard delivery separately from automatic paste and emit a content-free, generation-owned outcome
- show the existing transient, accessible Command-V cue for every confirmed clipboard-only dictation path
- reject duplicate, stale, delayed, and reordered lifecycle events without exposing transcript or clipboard content
- fence the queued main-thread clipboard/paste mutation against cancellation and supersession
- fail closed when either overlay event subscription cannot be established
- document the delivery event, retry result, and overlay contract

## Validation

- Fleet Mac mini, exact SHA `302074e04b8905215273ac3c0c30a859f4a0d1d4`: TypeScript; 864 frontend tests; rustfmt; cargo check; 928 Rust unit tests plus capture-helper, sidecar, transcription, and transform integration suites; strict Clippy; reference-doc validation
- Native macOS smoke at the same SHA: built the isolated `Local Dictation Dev.app` and DMG; launched the app against isolated dev identity/data; verified the main UI and notch overlay, capture helper startup, installed Core ML model warm-up to Ready, exact overlay placement, and clean startup logs; then quit the dev app
- Independent Rust and frontend reviews, including re-review of the queued-delivery fence and listener failure paths: no blockers

The bundle command produced the expected unsigned-updater warning after successfully creating the app, DMG, and updater archive because no release signing key was supplied.

## Murmur Bench

Murmur Bench: N/A — presentation and delivery-status signaling plus generation ownership hardening only; recognized text, delivered text for current recordings, and paste execution are unchanged.

Closes #571


georgenijo:
Addressed all review findings in `302074e0`:

- queued injection now rechecks current/non-cancelled/Processing ownership under the dictation mutex and retains that fence through clipboard/paste mutation; invalidated work returns a cancellation-shaped result with empty text
- both overlay subscriptions handle rejected registration promises and fail closed through a shared gate; parameterized tests cover each failure
- retry documentation now distinguishes `ClipboardOnly(PasteFailed)` from clipboard-write `Err`

Independent Rust/frontend re-review found no blockers. Exact-SHA Mac validation is green (864 frontend tests, 928 Rust tests plus integration suites, cargo check, rustfmt, strict Clippy, and reference docs). Native exact-SHA rebuild/launch smoke is in progress.

